Issue #37, opened 2014-11-09
Images created with --release stable won't boot.


adrelanos:
On a standard (no `grml-debootstrap` involved in creation) `wheezy` system, file exists:
```
/boot/grub/video.lst
```

On a standard (no `grml-debootstrap` involved in creation) `jessie` system, file exists:
```
/boot/grub/i386-pc/video.lst
```

While attempting to build a vmimage with grml-debootstrap using `--release stable` on Debian `wheezy`, I am seeing this in the log.

```
+ chroot /mnt/debootstrap.22326 update-grub
Generating grub.cfg ...
cat: /boot/grub/video.lst: No such file or directory
done
```

I think there is an issue with the following code block.

https://github.com/grml/grml-debootstrap/blob/master/grml-debootstrap#L1081

```
  case "$RELEASE" in
    lenny|squeeze|wheezy)
      cp "${MNTPOINT}"/usr/lib/grub/i386-pc/* "${MNTPOINT}/boot/grub/"
      ;;
    *)
      cp -a "${MNTPOINT}"/usr/lib/grub/i386-pc "${MNTPOINT}/boot/grub/"
      ;;
  esac
```

When someone uses `--release stable` then the "`lenny|squeeze|wheezy`" code path won't match. Therefore the "`*)`" code path will be used. Resulting in using the wrong path. (Confirmed this by looking into the image.) Therefore the image will not be bootable. This worked in earlier versions of grml-debootstrap.

Using `stable` as codename is not supported as per grml-debootstrap man page. So you could say, it's not a bug.

So in case you don't want to support `stable` as codename...

Could you perhaps break the build when using codename `stable` or `testing`? Do you want a pull request for this?

PR #43, opened 2014-11-29
Break when using generic codenames such as "stable" or "testing" because...

adrelanos:
... those are unsupported.

Fixes "Images created with --release stable won't boot." https://github.com/grml/grml-debootstrap/issues/37


mika:
Sorry for taking so long, merged with slightly adjusted commit message.Thanks!

Issue #22, opened 2014-07-13
better error handling


adrelanos:
`grml-debootstrap` often uses uses tools such as `cp` without checking their exit codes.

Would you be interested to get an early `trap ERR`, that checks for unexpected non-zero exit codes? In the few cases where we expect non-zero exit codes, those need to be overrules, but overall, I believe we could make `grml-debootstrap` more bug free.

Alternatively, we could add a lot more `eend $?`'s.

What do you think?

adrelanos:
Since pull request #51 was merged, this one can be closed.

PR #51, opened 2014-12-16
better error handler version2

adrelanos:
Implements #22 (better error handling). Hopefully addresses concerns from previous pull request "better error handler" (#31). This time it's a much less intrusive, smaller change.

Thanks for @hartwork for telling me about `set -E`.

```
export REPORT_TRAP_ERR='yes'
```

Reports unexpected non-zero exit codes.

```
export FAIL_TRAP_ERR='yes'
```

Fails on unexpected non-zero exit codes.

I created a VM image using both options. Congratulations. So far no unexpected non-zero exit codes. If this gets merged, I suggest to leave it disabled by default in short term and to enable `export REPORT_TRAP_ERR='yes'` as developer/tester. It will report unexpected non-zero exit codes but not annoy any further. Later when we're really sure to have all exit codes under control we can consider making `export FAIL_TRAP_ERR='yes'` the default.

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Filename:      /etc/debootstrap/chroot-script
 # Purpose:       script executed in chroot when installing Debian via grml-debootstrap
 # Authors:       grml-team (grml.org), (c) Michael Prokop <mika@grml.org>
@@ -8,6 +8,14 @@
 # GRML_CHROOT_SCRIPT_MARKER - do not remove this line unless you want to keep
 # this script as /bin/chroot-script on your new installed system
 ################################################################################
+
+# error_handler {{{
+if [ "$REPORT_TRAP_ERR" = "yes" ] || [ "$FAIL_TRAP_ERR" = "yes" ]; then
+   set -E
+   set -o pipefail
+   trap "error_handler" ERR
+fi
+# }}}
 
 . /etc/debootstrap/config    || exit 1
 . /etc/debootstrap/variables || exit 1
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -6,6 +6,38 @@
 # License:       This file is licensed under the GPL v2+
 ################################################################################
 
+# error_handler {{{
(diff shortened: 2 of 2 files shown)


mika:
Sorry for late reply: I'd like to wait until Debian/jessie has been released to keep the diff between our trunk/master and the jessie release sensible and this error handling is a too intrusive change for the time being, I'll take a look at it again once Debian/jessie has been released. Thanks!

adrelanos:
What about adding a `development` git branch, where more intrusive changes like this can be merged? You could then merge the `development` branch into `master` once `jessie` has been released.

That way we would not be blocked by waiting for `jessie`. Otherwise providing more pull requests gets difficult without certainty if and/or how this gets merged are very difficult (potential merge conflicts).

mika:
Introducing a development branch is just moving efforts, since I'd have to rebase the develpment branch against master over time, instead of just rebasing the PR once it's relevant. Sorry, I don't think it's worth the effort for such a small project. If you want to contribute further PRs and are unsure whether they might get included just talk to me in advance.

hartwork:
> this error handling is a too intrusive change

Could you elaborate how it is intrusive? I don't see how it is. The diff is rather small and without the environment variables being set, the patch should not change existing behavior. What am I missing?

mika:
It's e.g. modifying execution of the chroot script from /bin/sh to /bin/bash, this might be just fine, but such small changes often turn out to be problems in minor details. Currently I'm trying to get grml-debootstrap into shape for Debian/jessie and am just lacking time to include changes where I'm not 100% sure about the consequences yet.

hartwork:
I understand that after the freeze you do want to limit updates at Debian level to minimal patches for show stoppers. As a consequence, that should make #58, #59, #63 top priority for Jessie. Please confirm.

mika:
Yes, those are amongst the issues that I consider for inclusion in jessie, unsure what we might really get into jessie though.

adrelanos:
Now, that jessie is out, can this please be merged?

mika:
Jepp, in the works, let's git it some more testing - just merged. :) Thanks! :+1:

Issue #63, opened 2015-01-26
Git HEAD: SSH server process on host system stopped (rather than that inside the chroot)


hartwork:
Hello!

A friend of mine ran into a situation where grml-debootstrap killed the SSH server of the host system. @ZaphodB and I had a closer look at the code now. It seems that grml-debooststrap is stopping init scripts at two places: The `services` stage in file `chroot-script`
```
services() {
  for service in ssh mdadm mdadm-raid ; do
    if [ -x /etc/init.d/"$service" ] ; then
      /etc/init.d/"$service" stop || true
    fi
  done
}
```
and the `cleanup` function in file `grml-deboostrap`:
```
# make sure $TARGET is not mounted when exiting grml-debootstrap
if [ -n "$MNTPOINT" ] ; then
  if grep -q "$MNTPOINT" /proc/mounts ; then
    # make sure nothing is left inside chroot so we can unmount it
    [ -x "$MNTPOINT"/etc/init.d/ssh ] && "$MNTPOINT"/etc/init.d/ssh stop
    [ -x "$MNTPOINT"/etc/init.d/mdadm ] && "$MNTPOINT"/etc/init.d/mdadm stop
    ...
```
Now the latter is interesting.
With init script `/etc/init.d/ssh` (at least of Debian wheezy) command `stop` results in execution of
```
start-stop-daemon --stop --quiet --oknodo --pidfile /var/run/sshd.pid
```
with `/var/run/sshd.pid` being the host's file (since it was not run by/inside a chroot).
Essentially, `/etc/init.d/ssh stop` must not be called from outside the chroot.

We ran into this just once. It seems that `cleanup` needs to have been called from `bailout` (i.e. after some error) for this to trigger.

We are hoping for a soon fix. Thanks!

PR #71, opened 2015-03-10
Stop stopping host's (rather than chroot's) SSH and mdadm services (issue #63)

hartwork:


mika:
I think we should still cleanup the process list, chrooting into the environment should fix the reported behaviour. Just commited a change which is supposed to fix this as needed. Thanks for the PR!

hartwork:
For the record, @mika's re-made commit is f992b13b0e020e75dd4d6592f8d18f0ccafe9e5f.

Issue #55, opened 2015-01-07
Git HEAD: Claims to be the version installed system-wide (rather than the version in Git)


hartwork:
Rather than obtaining its version from `debian/changelog`, grml-debootstrap currently introduces itself to the user as the version installed system-wide, while being potentially different code.

How the current code works:
```
$ fgrep -w VERSION grml-debootstrap
VERSION="$(dpkg-query --show --showformat='${Version}' "$PN")"
VERSION="${VERSION:-unknown}"
  einfo "$PN - version $VERSION"
   einfo "$PN [${VERSION}] - Please recheck configuration before execution:"
```

An idea, for how it could work instead:
* Check for a folder `"$(dirname $(which "$0")")/.git"`
* If found, ask for `dpkg-parsechangelog | sed -e '/^Version: / !d' -e 's|^Version: ||'`
* Else, use `dpkg-query --show --showformat='${Version}' "grml-debootstrap"` as done at the moment

What do you think?

mika:
Fixed thanks to your PR, closing.

PR #73, opened 2015-03-10
Fixed version reported when run from Git (issue #55)

hartwork:
I refrained from using
```
dpkg-parsechangelog | sed -e '/^Version: / !d' -e 's|^Version: ||'
```
eventually, to not add a runtime dependency on `dpkg-dev`.
If that dependency is fine by you, feel free to replace
```
git describe | sed 's|^v||'
```
with the former.

Best, Sebastian

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -8,7 +8,11 @@
 
 # variables {{{
 PN="$(basename "$0")"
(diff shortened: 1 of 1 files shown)


mika:
Nice, thanks!

hartwork:
For record: picked commit was b3cddceda4de4d2417cbba3b7e8ae78fd72764f0.

Issue #78, opened 2015-04-29
Git HEAD: Installing jessie to LVM volume does not boot (due to grub-install failure)


hartwork:
Related to this bug in grub-install: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=783740 .
With grml-debootstrap, it looks like this:
```
[..]
   Executing stage grub_install
Installing grub on /dev/foo/bar:
Installing for i386-pc platform.
grub-install: error: disk `lvmid/I5GpDx-OoGE-BrAT-ik3g-ClI4-WT9H-YcTq5q/v7P1XE-ERW2-gIFB-oeut-4QNb-TgCg-yWeOmd' not found.
[..]
```

mika:
Merged PR, closing.

PR #81, opened 2015-05-06
Improve GRUB approach (issue #78)

hartwork:
PS: The readlink part fixes symlink issues with GRUB 1.99.


hartwork:
If this could be reviewed/merged before the weekend it would rock the house!

hartwork:
Found a bug with this, second take in a few minutes.

PR #50, opened 2014-12-16
Try unmounting [..]/dev harder (issue #41)

hartwork:
Improved version of proposed fix to issue #41. Please consider for application :)

PR #82, opened 2015-05-06
Fix unmounting [..]/dev once more (issue #50)

hartwork:


babilen:
It would be quite nice if this could be merged. Any particular reason why it isn't?

PR #83, opened 2015-05-06
Revert mistaken dependency

hartwork:
Previously, I overlooked that grub-mkimage is called from inside the chroot, only. Therefore it is not required on the host system.


mika:
Merged, thanks

PR #86, opened 2015-05-07
Improve GRUB approach (second take, issue #78)

hartwork:
The readlink part fixes symlink issues with GRUB 1.99.

If this could be reviewed/merged before the weekend it would rock the house!


mika:
Thanks, merged!

hartwork:
Awesome!

PR #89, opened 2015-06-17
Use official redirector address httpredir.debian.org

paulmenzel:
The address [`http.debian.net`](http://http.debian.net) is officially replaced by
[`httpredir.debian.org`](http://httpredir.debian.org).

Use the following command to update all places.

	git grep -l 'http.debian.net/debian' | xargs sed -i 's,http.debian.net/debian,httpredir.debian.org/debian,g'


mika:
Thanks!

PR #95, opened 2016-12-15
Export TARGET_HOSTNAME before running post-scripts

sts:
Export $HOSTNAME in $TARGET_HOSTNAME so one is able to do customize things based upon the hostname of the installed machine (eg. to populate /e/n/interfaces by doing a external database lookup).


mika:
Thanks for the PR :)

Issue #87, opened 2015-05-30
grub root=/dev/mapper/loop0p1


koppi:
I'm running grml-debootstrap as follows:

```bash
RELEASE=jessie
        
sudo http_proxy=http://localhost:3142/ \
     grml-debootstrap \
     --release $RELEASE \
     --hostname machinekit \
     --bootappend "xeno_hal.smi=1 lapic=notscdeadline hpet=disable i915.i915_enable_rc6=0 i915.powersave=0 intel_idle.max_cstate=1 processor.max_cstate=1 isolcpus=1 idle=poll" \
     --mirror http://ftp.de.debian.org/debian \
     --vmfile --target machinekit-$RELEASE.img \
     --vmsize 4G \
     --scripts $RELEASE/scripts \
     --chroot-scripts $RELEASE/chroot-scripts \
     --force \
     --password mk \
     --nokernel \
     --defaultinterfaces
```
When I boot the image in qemu, there is ```root=/dev/mapper/loop0p1``` and the kernel does not boot
![20150530-001](https://cloud.githubusercontent.com/assets/9458631/7897027/33cd4716-06cd-11e5-9c49-83c34923e561.png)

When I set ```root=/dev/sda1``` during grub boot, the kernel boots, when I run ```sudo update-grub``` inside the qemu instance, the entry becomes ```root=UUID=...``` and the image boots.

koppi:
What needs to be changed, that grml-debootstrap generates an entry with ```root=UUID=...```?

koppi:
```bash
$ dpkg -l | grep grml-debootstrap
ii  grml-debootstrap                                            0.69                                       all          wrapper around debootstrap for installing pure Debian
```

mika:
Wanted to reproduce this, but sadly ran into another GRUB bug (unsure whether this a known one yet): http://michael-prokop.at/screeni/screenshot.2015-06-03T17:56:45.png ->

````
source /boot/grub/grub.cfg
error: file `/boot/grub/fonts/unicode.pf2' not found.
error: no suitable video mode found.
grub> ls /boot/grub/
unicode.pf2 i386-pc/ grub.cfg
````

mika:
The mentioned unicode.pf issue is https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=787685

I'm not able to reproduce this bug with grml-debootstrap v0.70.0 executing:

````
% sudo grml-debootstrap \               
     --release jessie \
     --hostname grml \
[...]

PR #96, opened 2016-12-18
fix GRUB root=UUID=... for jessie, see grml/grml-debootstrap#87

hex2a:
see https://github.com/grml/grml-debootstrap/issues/87

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1289,7 +1289,7 @@ fi
   esac
 
   case "$RELEASE" in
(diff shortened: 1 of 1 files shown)


mika:
Sorry for the delay in handling this. I just took a closer look at it and indeed, we should *always* honor the GRUB_DISABLE_LINUX_UUID=true setting. I committed 900df89874f060271a621a5937ab8030df403419 with credits to you, thanks for your PR!

PR #97, opened 2017-02-07
one package per line

adrelanos:
No changes besides only one package per line instead of some lines with multiple packages per line.

This should make cleaning up that file as well as documenting this file in future easier.


mika:
Great idea, sorry for taking so long to merge this one! :+1:

PR #101, opened 2017-07-09
Update /dev and /dev/pts mount from --bind to proper fs type

crisidev:
We are using grml-debootstrap to create various virtual machine in parallel. Our build system started failing after updating to debian stretch, where grml-debootstrap version is 0.78.

When you run multiple grml-debootstrap in parallel, specifying a target in /dev, the bind mount of /dev and /dev/pts makes the successive umount and delete of the temporary mount folder failing.

I think this is caused by how --bind works with recent versions of systemd and udev.

Using proper filesystem definitions in the mount part and removing the bind mount completely solves our problem.

Is this something you think can be pushed upstream or could break compatibility on other systems which are not Debian based?

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1241,8 +1241,8 @@ finalize_vm() {
 
   mount -t proc none "${MNTPOINT}"/proc
   mount -t sysfs none "${MNTPOINT}"/sys
-  mount --bind /dev "${MNTPOINT}"/dev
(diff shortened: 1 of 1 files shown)


crisidev:
I worked on this change with @uovobw, @mindolo, @lcodeca.

mika:
Nice find, thanks a lot for the fix! 👍

PR #109, opened 2018-07-24
Initial VM test builds on TravisCI

mika:
We're starting a Docker container with Debian/stretch, installing
the Debian package build result (generated via travis.debian.net)
of grml-debootstrap inside it. Then execute grml-debootstrap,
installing to a VM file. Finally boot the generated VM file via
qemu.

We use stdio of qemu's serial console to check for the boot prompt
of the installed Debian system, using goss for running the actual
checks.

See https://travis-ci.org/mika/grml-debootstrap/jobs/407635286 for the actual test run on TravisCI.

### DIFF
--- a/travis/build-vm.sh
+++ b/travis/build-vm.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+TARGET="${TARGET:-/code/qemu.img}"
+RELEASE="${RELEASE:-stretch}"
+
+cd "$(dirname "$TARGET")"
+apt update
+apt -y install ./grml-debootstrap*.deb
+
+grml-debootstrap \
+  --force \
+  --vmfile \
+  --vmsize 3G \
+  --target "$TARGET" \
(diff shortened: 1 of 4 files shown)


travis/build-vm.sh
```diff
@@ -0,0 +1,21 @@
...
+grml-debootstrap \
+  --force \
+  --vmfile \
+  --vmsize 3G \
+  --target "$TARGET" \
+  --bootappend "console=ttyS0,115200 console=tty0 vga=791" \
+  --password grml \
+  --release  "$RELEASE" \
+  --hostname "$RELEASE" \
+  # EOF
```
zeha: the EOF line is a bit weird.

mika: Agreed, JFTR: it avoids having to keep the last line of actual code without `\`, so e.g. when adding a new cmdline option to the script or sorting lines later on then getting a possibly broken command line because of too much/too few `\` :))

PR #111, opened 2018-07-24
Improve checks to make sure loop and dm-mod module are present

jkirk:
This fix was triggered because running --vmfile to create a virtual
machine image file caused the following error:

| device-mapper: reload ioctl on loop1p1p1 failed: Device or resource busy
| create/reload failed on loop1p1p1

This was caused because kpartx was run twice. The first run was used to test if
the Device-mapper was ready and the second run was actually add the partition
mappings. This used to fail under some circumstances.

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1387,6 +1387,33 @@ prepare_vm() {
      bailout 1
   fi
 
+  # make sure loop module is present and an usable loop device exists
+  if ! losetup -f >/dev/null 2>&1; then
+    einfo "Can not find a usable loop device, retrying after loading loop module."
+    if ! modprobe loop > /dev/null 2>&1; then
+      eerror "Error loading loop module" ; eend 1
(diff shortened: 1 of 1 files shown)


zeha:
Maybe you want to opportunistically load all relevant modules before trying other commands with `modprobe -q -a dm-mod loop`. That would cut down on the LOC...

jkirk:
Actually I tried that, but on a Grml system I can not test without loop, and on my machine I can not test without dm-mod. So this is the best I came up with.

If you can provide a test system without a running loop device and device mapper I would be more than happy to shorten the LOC. But for now, I would keep it like it is...

jkirk:
Yes, please do so. Feel free to adapt the code if you like...

zeha:
pushed something that works for me

mika:
Pushed with one further commit on top, thanks folks! :)

PR #112, opened 2018-07-25
Improve cleanup of loop devices

jkirk:
kpartx (sometimes) does not cleanup the loop devices when deleting the
partition devmappings. This is a known bug, see [Debian Bug #891077.](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=891077)

This is change is a possible workaround for that issue.

Now aborting a grml-debootstrap is possible without leaving unneded loop
devices behind.

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -298,7 +298,13 @@ cleanup() {
 
   if [ -n "${ORIG_TARGET}" ] ; then
     einfo "Removing loopback mount of file ${ORIG_TARGET}."
-    kpartx -d "${ORIG_TARGET}" ; eend $?
+    kpartx -d "${ORIG_TARGET}"
+    # Workaround for a bug in kpartx which doesn't clean up properly,
+    # see Debian Bug #891077 and Github-PR grml/grml-debootstrap#112
+    if dmsetup ls | grep -q "^${LOOP_PART} "; then
+      kpartx -d "/dev/${LOOP_DISK}" >/dev/null
+    fi
+    eend $?
   fi
 }
 
@@ -1432,6 +1438,7 @@ prepare_vm() {
   # hopefully this always works as expected
   LOOP_PART="${DEVINFO##add map }" # 'loop0p1 (254:5): 0 20477 linear 7:0 3'
   LOOP_PART="${LOOP_PART// */}"    # 'loop0p1'
+  LOOP_DISK="${LOOP_PART%p*}"      # 'loop0'
   export TARGET="/dev/mapper/$LOOP_PART" # '/dev/mapper/loop1p1'
 
   if [ -z "$TARGET" ] ; then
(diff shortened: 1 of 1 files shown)


grml-debootstrap
```diff
@@ -298,7 +298,12 @@ cleanup() {
 
   if [ -n "${ORIG_TARGET}" ] ; then
     einfo "Removing loopback mount of file ${ORIG_TARGET}."
-    kpartx -d "${ORIG_TARGET}" ; eend $?
+    kpartx -d "${ORIG_TARGET}"
+    if dmsetup ls | grep -q ${LOOP_DISK}; then
```
mika: shellcheck is right, the variable should be quoted :) See https://travis-ci.org/grml/grml-debootstrap/jobs/408132398

grml-debootstrap
```diff
@@ -1516,6 +1522,9 @@ fi
   try_umount 3 "${MNTPOINT}"/dev
   umount "${MNTPOINT}"
   kpartx -d "${ORIG_TARGET}" >/dev/null
+  if dmsetup ls | grep -q ${LOOP_DISK}; then
```
mika: same here

grml-debootstrap
```diff
@@ -1432,6 +1437,7 @@ prepare_vm() {
   # hopefully this always works as expected
   LOOP_PART="${DEVINFO##add map }" # 'loop0p1 (254:5): 0 20477 linear 7:0 3'
   LOOP_PART="${LOOP_PART// */}"    # 'loop0p1'
+  LOOP_DISK="${LOOP_PART%p1}"    # 'loop0'
```
mika: This might break in the future if we're not installing to first partition? We could consider using "%p*" or so instead?

jkirk:
This time you got the code update before I was able to comment.. :)

grml-debootstrap
jkirk: Yeah, saw that and should be fixed in the following commit.

grml-debootstrap
jkirk: I would like to use "%p[0-9]" (for now) because I usually avoid "*" globbing. But if you think we ever will user double digit partitions I am willing to change that.

grml-debootstrap
jkirk: Ack and done.

jkirk:
After talking to @mika, I will squash my last commit.

jkirk:
Rebased to current master and squashed the (now missing) commit, see: https://github.com/grml/grml-debootstrap/pull/112#pullrequestreview-140707748

mika:
Did you forgot to squash d80bc9255f84663385b019c36cf077b141559de1 + 8a6b0ea698ebf68ecb45a0b66990a1202e37f5b3? Then I'm happy to do a `git merge --no-ff origin/pr/112` :)

jkirk:
Actually I did not "forget" to squash the commits. I wanted to keep the second one to make it visible that the issue was detected by spellsheck via the cool new Travis CI integration. But anyway, I squashed it now.

mika:
Thanks :)

PR #113, opened 2018-08-27
allow skip installation of grub

adrelanos:
Fully untested.

Rationale: in short: there is no grub-pc / grub for arm64 / Rasberry PI (RPI))
@Algernon-01 could maybe explain the rationale more elaborate if desired?

Testing and request for merge later unless someone is faster than we are.
Help welcome.

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -80,6 +80,7 @@ MNTPOINT="/mnt/debootstrap.$$"
 [ -n "$TUNE2FS" ] || TUNE2FS='tune2fs -c0 -i0'
 [ -n "$UPGRADE_SYSTEM" ] || UPGRADE_SYSTEM='yes'
 [ -n "$VMSIZE" ] || VMSIZE="2G"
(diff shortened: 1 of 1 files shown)


grml-debootstrap
```diff
@@ -1441,11 +1442,14 @@ prepare_vm() {
...
 # }}}
 
-# make VM image bootable and unmount it {{{
-finalize_vm() {
+# make VM image bootable {{{
+grub_install() {
   if [ -z "${VIRTUAL}" ] ; then
      return 0
   fi
+  if [ ! "$GRUB_INSTALL" = "yes" ]; then
```
mika: I'd propose to switch this into (please also notice the ${...} to stay consistent with the style in the function):

```
if [ "${GRUB_INSTALL}" != "yes" ] ; then
```

and also consider mentioning that we're skipping it because of GRUB_INSTALL, like:

```
einfo "Not installing GRUB as requested via \$GRUB_INSTALL=$GRUB_INSTALL"
```

PR #119, opened 2019-01-21
Add support for mounting /boot

paulmenzel:


paulmenzel:
This still needs to be tested, and should be considered as an RFC.

paulmenzel:
Creating the FS on the block device, and handling, if it’s a directory, is still missing.

grml-debootstrap
```diff
@@ -1385,6 +1391,26 @@ mount_target() {
...
+mount_boot_target() {
+  if [ -n "$MOUNT_BOOT" ] ; then
+     einfo "/boot is not on a separate partition, nothing to mount."
+  else
+     if grep -q "$MOUNT_BOOT" /proc/mounts ; then
+        ewarn "$MOUNT_BOOT already mounted, continuing anyway." ; eend 0
+     else
+       if ! [ -d "${MNTPOINT}/boot" ] ; then
+          [ -n "$VIRTUAL" ] || mkdir -p "${MNTPOINT}/boot"
+       fi
```
mika: Those three lines feel unnecessary, given that /boot should exist anyway and that you're creating it in line 1406 anyway :)

grml-debootstrap
```diff
@@ -1385,6 +1391,26 @@ mount_target() {
...
+       einfo "Mounting $MOUNT_BOOT to $MNTPOINT/boot"
+       mkdir -p "$MNTPOINT/boot"
+       mount -o rw,suid,dev "$MOUNT_BOOT" "$MNTPOINT/boot"
```
mika: Are you sure we need suid + dev permissions on /boot?
Also we should fail better with an actual error message.

grml-debootstrap
```diff
@@ -1385,6 +1391,26 @@ mount_target() {
 }
 # }}}
 
+# mount the new partition or if it's a directory do nothing at all {{{
+mount_boot_target() {
+  if [ -n "$MOUNT_BOOT" ] ; then
```
mika: Isn't this the wrong logic (as in: shouldn't this be -z ... here)?
Also we should consider verifying it's actually a valid block/LVM/... device, or at least that it can be accessed.
Since we don't format it on our own we should also check for a valid/supported filesystem?

While at it, please use "${MOUNT_BOOT}" instead of "$MOUNT_BOOT" etc. and two-space indention everywhere, that's the coding style we try to follow nowadays (yes, I'm aware that this is still inconsistent in our scripts :-/).

grml-debootstrap
```diff
@@ -1642,6 +1668,7 @@ preparechroot() {
   [ -n "$TARGET" ]              && echo "TARGET='$(sed "s,','\\\\'',g" <<<"${TARGET}")'"                           >> "$CHROOT_VARIABLES"
   [ -n "$UPGRADE_SYSTEM" ]      && echo "UPGRADE_SYSTEM='$(sed "s,','\\\\'',g" <<<"${UPGRADE_SYSTEM}")'"           >> "$CHROOT_VARIABLES"
   [ -n "$TARGET_UUID" ]         && echo "TARGET_UUID='$(sed "s,','\\\\'',g" <<<"${TARGET_UUID}")'"                 >> "$CHROOT_VARIABLES"
+  [ -n "$BOOT_UUID" ]           && echo "BOOT_UUID='$(sed "s,','\\\\'',g" <<<"${BOOT_UUID}")'"                     >> "$CHROOT_VARIABLES"
```
mika: This is unused and not yet handled inside chroot-script?

grml-debootstrap
```diff
@@ -372,6 +374,9 @@ while :; do
   --target|-t)         # Target partition (/dev/...) or directory
     shift; _opt_target="$1"
     ;;
+  --mountboot)         # Target partition (/dev/...) or directory
+    shift; _opt_mountboot="$1"
```
mika: I think you forgot to add it to the getopt handling.

grml-debootstrap
```diff
@@ -1385,6 +1391,26 @@ mount_target() {
 }
 # }}}
 
+# mount the new partition or if it's a directory do nothing at all {{{
+mount_boot_target() {
+  if [ -n "$MOUNT_BOOT" ] ; then
+     einfo "/boot is not on a separate partition, nothing to mount."
```
mika: You can just "return 0" afterwards and don't need the `else ... ` afterwards, since nothing from this function needs to be executed if separate /boot isn't requested.

mika:
@paulmenzel ping, are you still interested in getting this submitted? :)

jkirk:
I stumbled over this today, so it would be really nice if we can get this into grml-debootstrap. :)

PR #122, opened 2019-01-25
Replace http with https where possible

adrelanos:


mika:
thanks for spotting this, but I'd like to better match this with 7827bce99e0be17690494ec747f0b7875fc77e32 and will adjust it accordingly, thx!

PR #123, opened 2019-01-25
use debian https mirror by default

adrelanos:


zeha:
Thats not going to work on a number of (older) distributions.

mika:
> Thats not going to work on a number of (older) distributions.

ACK, and I think it also depends on apt-transport-https being present (at least for certain apt versions)?

mika:
Quoting from https://wiki.debian.org/SourcesList:

> The security.debian.org hosts currently do not have publicly verifiable SSL certificates on HTTPS and hence cannot be used with HTTPS at the moment

So the benefit is quite small overall anyway, at least until security.debian.org is available via https as well? Also we would like to have per-distribution support for the package list file, so we install apt-transport-https only for Debian releases being stretch or older, and also install the package only if it's relevant to the selected mirror. (apt-transport-https also pulls in libcurl3-gnutls, from source package curl which has quite some amount of security updates, see https://security-tracker.debian.org/tracker/source-package/curl).

PR #124, opened 2019-01-25
use https by default

adrelanos:


mika:
This doesn't work at all AFAICT as security.debian.org isn't available behind https.
Did you test this?

adrelanos:
Michael Prokop:
> This doesn't work at all AFAICT as security.debian.org isn't available behind https.
> Did you test this?


https should work.

This is what is being used in Whonix by default:

cat /etc/apt/sources.list.d/debian.list | grep --invert-match "#"

deb tor+https://deb.debian.org/debian-security/ buster/updates main
contrib non-free
deb tor+https://deb.debian.org/debian buster main contrib non-free

tor+ can also be dropped.

PR #138, opened 2019-08-21
travis: default to buster and also use debian:buster as docker container

mika:


mika:
Replaced with https://github.com/grml/grml-debootstrap/pull/143

PR #140, opened 2019-10-15
/etc/fstab: provide header comment, pointing to man pages and systemctl daemon-reload

mika:
Inspired by Marc Haber's bug report #936050 and what Fedora does with their anaconda installer.


mika:

mika:
Thanks for your reviews, @zeha!

PR #141, opened 2019-10-15
Properly exit if GRUB installation fails

mika:
For example when partition table is GPT but grub installation
expects a MS-DOS partition table, then grub-install fails
but grml-debootstrap ignored this. Since the system isn't bootable
then this might go unnoticed, so properly fail as soon as grub-install
returns with exit code != 0.

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -616,19 +616,31 @@ grub_install() {
      for device in $SELECTED_PARTITIONS ; do
         GRUB="${device%%[0-9]}"
         echo "Installing grub on ${GRUB}:"
-        grub-install --no-floppy "$GRUB"
+        if ! grub-install --no-floppy "$GRUB" ; then
+          echo "Error: failed to execute 'grub-install --no-floppy $GRUB'." >&2
+          exit 1
+        fi
+
      done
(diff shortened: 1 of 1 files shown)


mika:

chroot-script
```diff
@@ -616,19 +616,31 @@ grub_install() {
...
+
      done
      rm -f /boot/grub/device.map
   else
      echo "Installing grub on ${GRUB}:"
      case "$RELEASE" in
        lenny|squeeze|wheezy)
-         grub-install --no-floppy "$(readlink -f "${GRUB}")"
+         local grub_dev
+         grub_dev="$(readlink -f "${GRUB}")"
```
zeha: not sure about the quoting here?

mika: You mean because of the `"${GRUB}"` inside `"$(...)"`. If so, that's expected as such, demonstration:

```
% cat foo 
#!/bin/bash
DEV="/dev/foo bar"
foo "$(readlink -f ${DEV})"
% shellcheck foo

In foo line 3:
foo "$(readlink -f ${DEV})"
                   ^-- SC2086: Double quote to prevent globbing and word splitting.
```

Issue #106, opened 2017-10-17
BOOT_APPEND works only for VMs


kivoli:
Appending stuff to the Kernel cmdline via `BOOT_APPEND` in the config file or `--bootappend` on the CLI does not work because the responsible code is only executed for virtual machines:

https://github.com/grml/grml-debootstrap/blob/351237a25834610576f953d1be2976f043e6f862/grml-debootstrap#L1278

# Steps to reproduce the problem

1. Prepare physical machine for debootstrap installation
2. Execute `grml-debootstrap` with `--appendboot 'net.ifnames=0 biosdevname=0'`

# Expected behavior

Content of `BOOT_APPEND` appended to `GRUB_CMDLINE_LINUX_DEFAULT` in `/etc/default/grub`.

# Actual behavior

Does not happen.

# Your system configuration

Physical hardware

# Software versions

grml-debootstrap 0.78 (from inside grml 2017.05 Freedatensuppe)

suntong:
+1 -- I'm just about to report the same thing.

grml-debootstrap from latest grml-small, v0.85.

mika:
I started to work on this (completely untested though), so if anyone is interesting in this feature and/or can help us in testing, please take a look at https://github.com/grml/grml-debootstrap/pull/137

suntong:
I tested and can confirm that the issue has been fixed in `grml-debootstrap - version 0.89~bootappend01`

mika:
Thanks for testing this, @suntong - appreciated!
I just submitted this via https://github.com/grml/grml-debootstrap/pull/142

If anyone notices any problems please let us know, thanks!

PR #142, opened 2019-10-15
Support BOOT_APPEND usage outside of VMs

mika:
BOOT_APPEND works only for VMs when being invoked within
grub_install() inside main grml-debootstrap, move its
handling to chroot-script instead.

See https://github.com/grml/grml-debootstrap/issues/106


mika:

PR #143, opened 2019-10-15
travis: default to buster and also use debian:buster as docker container

mika:
The unstable environment seems to be flasky, though the
stretch one seems to be stable so far, let's try to update
it towards buster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
   - docker
 
 env:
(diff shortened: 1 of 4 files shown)


mika:
So travis only fails for the unstable build, unsure why that's so, but this shouldn't be a showstopper from my PoV, looking at https://travis-ci.org/grml/grml-debootstrap/builds this seems to be flaky so it's worth merging this

PR #177, opened 2021-06-03
config: Replace em dash — by `--` for switch in comment

paulmenzel:
Fixes: d88d27d7 ("drop enabled config variables from config file and use defaults")


mika:
Thanks, merged!

PR #182, opened 2021-06-14
Implement option vmtype to select between raw or qcow2

paulmenzel:


mika (commented):
I'm not yet convinced that we should support this from within grml-debootstrap, tbh. :)

You write:

> It looks like though, installing the system into that file using NBD is twice as slow.

So what's the actual reason, to not install it into raw format as-is, and then convert to the format as needed?

Also all the `qemu-nbd` actions around `/dev/nbd0` seem to ask for trouble AFAICS, what about e.g. if the device is used elsewhere already? :-/
(Not commenting on other code changes themselves yet)

paulmenzel:
> I'm not yet convinced that we should support this from within grml-debootstrap, tbh. :)
> 
> You write:
> 
> > It looks like though, installing the system into that file using NBD is twice as slow.
> 
> So what's the actual reason, to not install it into raw format as-is, and then convert to the format as needed?

I only came up with, that on the “build machine” you not have 100 GB of space for the raw image.

> Also all the `qemu-nbd` actions around `/dev/nbd0` seem to ask for trouble AFAICS, what about e.g. if the device is used elsewhere already? :-/
> (Not commenting on other code changes themselves yet)

Yes, I have to figure out – the same with kpartx – how to get the correct device.

grml-debootstrap
```diff
@@ -308,14 +310,19 @@ cleanup() {
...
+      if dmsetup ls | grep -q "^${LOOP_PART} "; then
+        kpartx -d "/dev/${LOOP_DISK}" >/dev/null
+      fi
+      eend $?
+    else
+      einfo "Removing loopback mount of file ${ORIG_TARGET}."
+      qemu-nbd -d /dev/nbd0
     fi
-    eend $?
   fi
```
jkirk, line 326: @paulmenzel Just a quick question: you moved `eend $?`. Any reason for not keeping it at the end?

paulmenzel: Thank you. That’s an oversight no my part.

mika:
Hm, what do we want to do about this one? Do maybe @adrelanos or @ArrayBolt3 have any opinion about this? :thinking: 
IMO we should avoid adding more and more options to grml-debootstrap and instead make it use good defaults.

ArrayBolt3:
> I only came up with this, because on the “build machine” you do not always have 100 GB of free space for the raw image.

If you're using a good Linux filesystem like ext4, qemu-img (and therefore grml-debootstrap) should make a sparse file for the VM image. Even if you make a 100GB image, it will only take up as much space as it needs. You might need to be careful when copying the file around (i.e. use `cp --sparse=always`), but it shouldn't take up 100GB at once.

mika:
> > I only came up with this, because on the “build machine” you do not always have 100 GB of free space for the raw image.
> 
> If you're using a good Linux filesystem like ext4, qemu-img (and therefore grml-debootstrap) should make a sparse file for the VM image. Even if you make a 100GB image, it will only take up as much space as it needs. You might need to be careful when copying the file around (i.e. use `cp --sparse=always`), but it shouldn't take up 100GB at once.

Thank you for your reply! I take this as a "no need for this change"? :smiley:

ArrayBolt3:
I do think it's probably not needed, yes. It adds more complexity, and if a true qcow2 image is needed it's easy enough to convert a raw image to qcow2. That will take extra disk space unfortunately though, so there might be an edge case where this would be helpful, but I'm not sure the added complexity and possible danger are worth it.

mika:
I see, thanks for clarifying this also from your PoV and being so responsive, @ArrayBolt3 - appreciated! :bow: 
Thanks also @paulmenzel for providing the PR, sorry that it didn't work out at the end but I think everyone learned from it, looking forward to further contributions of yours! :+1:

Issue #185, opened 2021-08-08
Unaligned partitions with `--vm` and device target


lxp:
I am running Debian bullseye with grml-debootstrap 0.96 and parted 3.4-1.
When running with `--vm` and a device file as target, unaligned partitions are created.

This seems to be caused by the non-optimal start and end arguments used in the mkpart command:

https://github.com/grml/grml-debootstrap/blob/86f83aa4b5047fafa74ef003d2da70bf1eeb50a7/grml-debootstrap#L1477

I tried to analyse it further by manually running parted and fdisk:

```sh
$ sudo parted /dev/vgcrypt/test
GNU Parted 3.4
Using /dev/dm-4
Welcome to GNU Parted! Type 'help' to view a list of commands.
(parted) mklabel msdos
(parted) mkpart primary ext4 2M -1
Warning: The resulting partition is not properly aligned for best performance: 3906s % 4096s != 0s
Ignore/Cancel? i
(parted) print
Model: Linux device-mapper (linear) (dm)
Disk /dev/dm-4: 21,5GB
Sector size (logical/physical): 512B/512B
Partition Table: msdos
Disk Flags: 

Number  Start   End     Size    Type     File system  Flags
 1      2000kB  21,5GB  21,5GB  primary  ext4         lba

(parted)

$ sudo fdisk /dev/vgcrypt/test                           

Welcome to fdisk (util-linux 2.36.1).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.


Command (m for help): p
Disk /dev/vgcrypt/test: 20 GiB, 21474836480 bytes, 41943040 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 524288 bytes / 2097152 bytes
Disklabel type: dos
Disk identifier: 0xf08bb2c7

Device             Boot Start      End  Sectors Size Id Type
/dev/vgcrypt/test1       3906 41941087 41937182  20G 83 Linux

Partition 1 does not start on physical sector boundary.
```

The start `2M` is interpreted as 2 Megabyte (2000000 bytes or 3906 sectors) from the disk start.
The end `-1` is interpreted as 1 Megabyte (1000000 bytes or 1953 sectors) before the disk end.
I suspect there was some change in the alignment handling or constraint solving between parted 3.2 and parted 3.4, but I could not find the actual cause, why it behaves differently.
[...]

PR #190, opened 2021-12-06
VMs: create partition with alignment starting at 4MiB + ending at 100%

mika:
Our partition creation process inside VMs used parted with "mkpart
primary ext4 2M -1", which could end up with unaligned partitions (as
reported in #185):

| (parted) mkpart primary ext4 2M -1
| Warning: The resulting partition is not properly aligned for best performance: 3906s % 4096s != 0s

This depends on the I/O settings of the underlying storage.

The mkpart start at "2M" is interpreted as 2 Megabyte (2000000 bytes or
3906 sectors) from the disk start.  The end "-1" is interpreted as 1
Megabyte (1000000 bytes or 1953 sectors) before the disk end.

On the other hand, "2MiB" is interpreted as 2 Mebibyte (2097152 bytes or
4096 sectors) from the disk start, and "100%" is interpreted as the disk
end.

Quoting from https://www.gnu.org/software/parted/manual/html_node/unit.html:

| Parted will compute sensible ranges for the locations you specify (e.g.,
| a range of +/- 500 MB when you specify the location in “G”, and a range
| of +/- 500 KB when you specify the location in “M”) and will select the
| nearest location in this range from the one you wrote that satisfies
| constraints from both the operation, the filesystem being worked on, the
| disk label, other partitions and so on.
| [...]
| Note that as of parted-2.4, when you specify start and/or end values
| using IEC binary units like “MiB”, “GiB”, “TiB”, etc., parted treats
| those values as exact, and equivalent to the same number specified in
| bytes (i.e., with the “B” suffix), in that it provides no “helpful”
| range of sloppiness. Contrast that with a partition start request of
| “4GB”, which may actually resolve to some sector up to 500MB before or
| after that point. Thus, when creating a partition, you should prefer
| to specify units of bytes (“B”), sectors (“s”), or IEC binary units
| like “MiB”, but not “MB”, “GB”, etc.

Furthermore quoting from https://www.gnu.org/software/parted/manual/html_node/mkpart.html#FOOT2:

| Cheap flash drives will be with us for a long time to come, and, for
| them, 1MiB alignment is not enough. Use at least 4MiB-aligned
| partitions. For details, see Arnd Bergman’s article,
| http://lwn.net/Articles/428584/ and its many comments.

A better default should be usage of an IEC binary unit with a
4MiB-aligned partition, so let's switch to "4MiB 100%".

Thanks: David Gnedt for the bug report and analysis, Darshaka Pathirana and Chris Hofstaedtler for further information and feedback
Closes: grml/grml-debootstrap#185

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1474,7 +1474,7 @@ prepare_vm() {
     dd if="${MBRTMPFILE}" of="${TARGET}" conv=notrunc
     eend $?
   fi
-  parted -s "${TARGET}" 'mkpart primary ext4 2M -1'
+  parted -s "${TARGET}" 'mkpart primary ext4 4MiB 100%'
   parted -s "${TARGET}" 'set 1 boot on'
 
   DEVINFO=$(kpartx -asv "$TARGET") # e.g. 'add map loop0p1 (254:5): 0 20477 linear 7:0 3'


jkirk (approved):

mika:
Thanks folks! :+1:

Issue #169, opened 2021-04-27
ARM support


adrelanos:
Could you please add support for creating VM images with ARM platform support?

mika:
More than happy to accept patches/PRs :)

GavinPacini:
@mika I'm happy to look into this. I've already taken the `grml-debootstrap` output and made a bootable arm64 image from it.

In regards to that, I'd love to have your input on my thinking. Basically, the main differences with arm64 are going to be the kernel and grub setup. 
 - Kernel: needs to be arm64
 - Grub: needs to use the UEFI boot pattern i.e. a separate EFI partition in the first few sectors of the image file. This means only GPT / EFI compatible images will be bootable when arch=arm64.

Are you okay if I basically switch on `--arch arm64` and then make the relevant changes using that?

mika:
@GavinPacini this sounds good for me! Feel free to also throw PoC/incomplete code into my/our direction if you'd like early feedback/code review!

PR #191, opened 2022-03-24
Initial arm64 support

GavinPacini:
Hi there, I finally decided to tackle #169, about time. :)

This is an initial implementation. I've tested it with:
```sh
sudo UPGRADE_SYSTEM='no' ./grml-debootstrap \
        	--debopt "--verbose" \
        	--arch arm64 \
        	--filesystem ext4 \
        	--hostname host \
        	--nopassword \
        	--nopackages \
        	--release bullseye \
        	--keep_src_list \
        	--verbose \
        	--vmfile \
        	--vmsize 15G \
        	--target ../debian-grml.raw
```
and I've been able to boot this image with qemu on an Apple Silicon based Mac (through UTM).

Trouble is that I cannot really make sure I didn't break the amd64 builds, albeit I made every effort not to. 

I plan to test it a bit more but would appreciate an initial review at this stage. Thanks!


GavinPacini:
Thanks @mika! No problem, appreciate any feedback once you have the chance.

mika:
@GavinPacini finally had some time to take a look at it. Overall the PR looks good to me, there are some minor details we could improve, but let's polish this once we know it's working as expected. :)

I tried installing it into a VM, running:

```
CONFFILES=$(pwd) ./grml-debootstrap --vm --vmfile --target /dev/shm/debian-arm64.img -r bullseye --arch arm64 --password grml
```

But this failed for me with:

```
However the following packages replace it:
  grub2-common grub-common

E: Package 'grub-pc' has no installation candidate
```

This is related to the usage of `--nopackages` (adding that option then built the VM as expected), so I suppose we need to further adjust `chroot-script` to properly handle the grub situation also for arm64? :)

GavinPacini:
Thanks for reviewing @mika! 

I just reproduced that myself, thanks for bringing it up. After some digging, `grub-pc` seems to be the standard way to setup MBR / BIOS systems. There is no real support for MBR / BIOS on arm64 architectures (they use UEFI only). For this reason, we use `grub-2-common` and `grub-efi-arm64` packages ([as shown in the diff here](https://github.com/grml/grml-debootstrap/pull/191/files#diff-566641cf801fd8006b3b94fb6b0e3fa53f9a8de4b21ee08969d9185d19e158ddR1560)).

Basically, arm64 images should never have `grub-pc` installed. However `grub-pc` is in the `packages` file. Once I removed this from the `packages` file it created the VM fine. I see three options:

1. Create a separate `packages` file for arm64 images (e.g. `packages-arm64`) which we can switch to based on the `--arch` flag.
2. Add some flags into the `packages` file which allows us to denote if a package should be skipped for certain architectures.
3. Hardcode in ignoring `grub-pc` for arm64 images.

I would probably pick number 1 as the most future proof, but there would be a lot of duplication for now.

What do you think?

adrelanos:
`1.` seems also the most reliable, least surprising to me.
`2.` seems rather complex to implement, not very intuitive.
`3.` would be the most surprising.

So From the three options, I like `1.` most.

fosple:
@GavinPacini Thanks for looking into it. Are you still on it? 😊

adrelanos:
Which of these options would be acceptable? @mika

GavinPacini:
Hey @fosple, once @mika provides some direction on the preferred options mentioned above, I am happy to finish it out.

mika:
Sorry for the delay, I also think that the first option (`Create a separate packages file for arm64 images (e.g. packages-arm64) which we can switch to based on the --arch flag.`) seems preferable given our current situation!

GavinPacini:
Hi again, I've just pushed something that I think solves the outstanding issues. @mika I would appreciate another review when you have a chance. Thanks!

mika:
@GavinPacini great, thanks! Your changes look good to me on a quick sight, though I need a more quiet time to give it a more detailed look. Also in Debian we're in freeze stage for bookworm (see https://release.debian.org/testing/freeze_policy.html), so it might make sense to apply those changes once we can upload towards Debian/unstable with changes like new features again. :)

grml-debootstrap
```diff
@@ -1306,6 +1311,15 @@ mkfs() {
   fi
 
   if [ -n "$MKFS" ] ; then
+
+    if [ -n "$ARM_EFI_TARGET" ] ; then
+      einfo "Running mkfs.vfat $MKFS_OPTS on $ARM_EFI_TARGET"
+      mkfs.vfat -n "EFI" $MKFS_OPTS "$ARM_EFI_TARGET"
```
paulbaumgart, line 1317: FYI, the inclusion of `$MKFS_OPTS` was causing an error for me after merging this branch into the latest master:

```
* Running mkfs.vfat  -O ^metadata_csum_seed on /dev/mapper/loop15p1
mkfs.fat 4.1 (2017-01-24)
mkfs.vfat: invalid option -- 'O'
```

Removing `$MKFS_OPTS` from this line got me past it. At a glance, all the $MKFS_OPTS seem to only apply to `mkfs.ext4`, so it might not be needed here at all.

GavinPacini: You're right, have pushed a commit which fixes this. Thanks!

GavinPacini:
@mika Hey, just wondering what else is needed to get this merged in? I just want to make sure we don't miss the next release window. Thanks!

mika:
@GavinPacini I'll try to look into this in the next few days to get this merged ASAP! Thanks for the ping

chroot-script
```diff
@@ -269,14 +269,21 @@ packages() {
...
   if [ "$PACKAGES" = 'yes' ] ; then
-     if ! [ -r /etc/debootstrap/packages ] ; then
-       echo "Error: /etc/debootstrap/packages (inside chroot) not found, exiting." >&2
-       exit 1
-     else
-       $APTUPDATE
-       # shellcheck disable=SC2086,SC2046
-       DEBIAN_FRONTEND=$DEBIAN_FRONTEND $APTINSTALL $(grep -v '^#' /etc/debootstrap/packages) $GRMLPACKAGES
-     fi
+      PACKAGES_FILE="/etc/debootstrap/packages"
```
mika, line 272: nitpick: I'd appreciate to use two-space indention also here, to match what we have elsewhere in the script

chroot-script
```diff
@@ -269,14 +269,21 @@ packages() {
...
+
+      if [ "$ARCH" = 'arm64' ]; then
+        PACKAGES_FILE="/etc/debootstrap/packages-arm64"
+      fi
+
+      if ! [ -r $PACKAGES_FILE ] ; then
```
mika, line 278: nit: "${PACKAGES_FILE}" (to have "${...}" style variable and double-quotes to not break if the file name should include white spaces)

grml-debootstrap
```diff
@@ -257,6 +257,10 @@ cleanup() {
     einfo "Removing ${STAGES}" ; rmdir "$STAGES" ; eend $?
   fi
 
+  if [ -n "$ARM_EFI_TARGET" ]; then
```
mika, line 260: nit: "${ARM_EFI_TARGET}"

grml-debootstrap
```diff
@@ -1511,14 +1539,38 @@ grub_install() {
     bailout 1
   fi
 
+  if [ -n "$ARM_EFI_TARGET" ]; then
```
mika, line 1547: nit: "${ARM_EFI_TARGET}" (same below in line 1555 etc)

grml-debootstrap
```diff
@@ -1570,6 +1622,8 @@ fi
 
   einfo "Updating grub configuration file."
   chroot "${MNTPOINT}" update-grub
+  chroot "${MNTPOINT}" sync
+  sleep 2
```
mika, line 1619: this one feels strange, what's the reason for the "sleep 2" here?

grml-debootstrap
```diff
@@ -1465,29 +1475,47 @@ prepare_vm() {
...
   fi
-  parted -s "${TARGET}" 'mklabel msdos'
-  if [ "$FIXED_DISK_IDENTIFIERS" = "yes" ] ; then
-    einfo "Adjusting disk signature to a fixed (non-random) value"
-    MBRTMPFILE=$(mktemp)
-    dd if="${TARGET}" of="${MBRTMPFILE}" bs=512 count=1
-    echo -en "\\x41\\x41\\x41\\x41" | dd of="${MBRTMPFILE}" conv=notrunc seek=440 bs=1
-    dd if="${MBRTMPFILE}" of="${TARGET}" conv=notrunc
-    eend $?
+
```
mika, line 1482: nit: doesn't seem to be visible in the web view here, but here we have a trailing white space char

grml-debootstrap
```diff
@@ -1582,7 +1636,7 @@ fi
   if grep -q '^GRUB_DISABLE_LINUX_UUID=.*true' "${MNTPOINT}"/etc/default/grub 2>/dev/null ; then
     ewarn "GRUB_DISABLE_LINUX_UUID is set to true in /etc/default/grub, not adjusting root= in grub.cfg."
     ewarn "Please note that your system might NOT be able to properly boot."
-  else
+  elif [ -z "$ARM_EFI_TARGET" ]; then
```
mika, line 1632: why do we skip this for arm64 then, is the `root=...` configuration really working as needed? *hmmm*

grml-debootstrap
```diff
@@ -1798,8 +1786,14 @@ preparechroot() {
...
-    cp $VERBOSE "${_opt_packages:-$CONFFILES/packages}" \
-      "${MNTPOINT}"/etc/debootstrap/packages
+    PACKAGES_FILE="packages"
+
+    if [ "$ARCH" = 'arm64' ]; then
+      PACKAGES_FILE="packages-arm64"
+    fi
+
+    cp $VERBOSE "${_opt_packages:-$CONFFILES/$PACKAGES_FILE}" \
+      "${MNTPOINT}"/etc/debootstrap/$PACKAGES_FILE
```
mika, line 1796: nit: "${MNTPOINT}/etc/debootstrap/${PACKAGES_FILE}"

mika (commented):
Overall LGTM, the main issue I'm struggling with is the "sleep 2", so I'd like to get rid of this. 

The rest are only nitpicks :) If you'd prefer me to take care of the nitpicks myself, let me know.

IMO we should squash all your commits into one single commit (I can also take care of this if you'd prefer me to do so), and then we should be ready for pushing a new release out :)

mika:
@GavinPacini let's try to get this merged before more changes pile up and cause further merge conflicts :) I took your PR and tried to get it ready for submission, please see https://github.com/grml/grml-debootstrap/pull/214 - feel free to take it over and force-push it here or so - all credits should be toally yours of course! :) (When reviewing, ignore all the indention related changes, e.g. via `git show --ignore-all-space` on the cmdline)

Disclaimer: _not_ yet tested/verified on my side. BTW, what's your environment to test this? An actual arm64 system, or a VM?

I've only two things in mind now:

1) why do we skip the `Adjusting grub.cfg for successful boot sequence.` in the arm64 situation? (Only trying to understand the situation we're in)
2) now also having EFI support for VMs (see f4d3906aae588d8dac1a146f05151211073cbb80), maybe there is something we could do to simplify code and unify code, see e.g.:

```
parted -s "${TARGET}" 'mkpart ESP fat32 1MiB 101MiB'
```

vs

```
parted -s "${TARGET}" 'mkpart EFI fat32 1MiB 10MiB'
```

But perfect is the enemy of good, I think we can further improve/iterate of anyway, though I'd appreciate your feedback before submitting this :)

adrelanos:
I also think it's better to merge imperfect stuff fast for the purpose of avoiding merge conflicts.

You can withhold it from user documentation and/or label it an alpha/unsupported feature meanwhile if not (yet) tested. Since this is primarily a sysadmin / developer tool, the facts can just be documented and/or mentioned by the scripts. By facts I mean mentioning "community supported architecture, untested".

With increasing platform support it won't be possible for the main developer(s) to keep all tested. If github actions supported arm64 (not at time of writing but planned according to the tickets), we could use that. Perhaps in the future.

mika:
FYI: if anyone of you can give https://github.com/grml/grml-debootstrap/pull/214 a try and let me know whether this works for you, I'd submit it ASAP.

adrelanos:
I tested it but a bit too slow. Before today's rebase.

I am not an expert for qemu but this is my quick way to test boot it.

```
#!/bin/bash

set -x
set -e

## We might get away with --no-install-recommends, fewer packages.
sudo apt-get install qemu-system-arm qemu-system-aarch64 qemu-efi-aarch64 qemu-utils

qemu-system-aarch64 \
    -M virt \
    -cpu cortex-a57 \
    -m 1024 \
    -bios /usr/share/qemu-efi-aarch64/QEMU_EFI.fd \
    -drive format=raw,file=/home/user/grml-debootstraptestbin/test.img \
    -nographic
```

Happy to report that build + boot was success.

adrelanos:
Continued testing here:
https://github.com/grml/grml-debootstrap/pull/214#issuecomment-1771021365

mika:
@GavinPacini we took care of your PR via https://github.com/grml/grml-debootstrap/pull/214, thanks for your work!

Issue #199, opened 2022-11-21
fstab has unknown errors option for xfs


zeha:
When installing with `--filesystem xfs`, the generated `/etc/fstab` still ends up with `errors=remount-ro` for `/`. The kernel in Debian bullseye does not understand this option, and thus mounts `/` as read-only.

(Remounting rw does not work, but this is just a follow up issue.)

PR #201, opened 2022-11-25
Use "errors=remount-ro" fstab option only within supported filesystems

mika:
Also XFS doesn't support the errors=remount-ro option, but instead of disabling the option for file systems that don't support it, let's only enable it for the once that support it (according to linux's source).

Thanks: Chris Hofstaedtler for the bug report
Closes: https://github.com/grml/grml-debootstrap/issues/199

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -479,11 +479,16 @@ createfstab(){
 EOF
 
   if [ -n "$TARGET_UUID" ] ; then
-    local rootfs_mount_options=",errors=remount-ro"
+    local rootfs_mount_options=""
+
+    if [ -z "${FILESYSTEM}" ] ; then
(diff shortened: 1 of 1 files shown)


chroot-script
```diff
@@ -479,11 +479,16 @@ createfstab(){
...
+
+    if [ -z "${FILESYSTEM}" ] ; then
+      FILESYSTEM="$(blkid -o value -s TYPE /dev/disk/by-uuid/"${TARGET_UUID}")"
+    fi
+
     case "${FILESYSTEM}" in
-      f2fs)
-        # errors=remount-ro is unsupported, see https://github.com/grml/grml-debootstrap/issues/163
-        rootfs_mount_options=""
+      # errors=remount-ro is supported only by some few file systems
```
zeha, line 489:
```suggestion
      # errors=remount-ro is supported only by a few file systems
```

mika: ACK, thx!

Issue #200, opened 2022-11-21
Installing on xfs should add xfsprogs


zeha:
When installing on an xfs root partition, xfsprogs should be installed.
Without it, `update-initramfs` complains:

```
update-initramfs: Generating /boot/initrd.img-5.10.0-19-amd64
W: /sbin/fsck.xfs doesn't exist, can't install to initramfs
```

PR #202, opened 2022-11-25
Ensure to have filesystem tools installed

mika:
When installing to an XFS root partition, update-initramfs complains:

| update-initramfs: Generating /boot/initrd.img-5.10.0-19-amd64 | W: /sbin/fsck.xfs doesn't exist, can't install to initramfs

Let's make sure to have the according filesystem package available.

FTR, no need for handling ext* + e2fsprogs, as this is available anyways due to `Priority: required`.

Also, the "--filesystem jfs" isn't supported by default (as mkfs.jfs expects user input unless invoked with -q, something we might wanna automate in the future?), but someone can prepare it manually and install it to the mounted file system, so let's also handle this.

Thanks: Chris Hofstaedtler for the bug report
Closes: https://github.com/grml/grml-debootstrap/issues/200

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -525,6 +525,28 @@ fstab() {
 }
 # }}}
 
+# ensure we have according filesystem tools available {{{
+install_fs_tools() {
+  local pkg=""
+
+  # note: this is supposed to be coming either via command lines'
+  # $_opt_filesystem or via createfstab()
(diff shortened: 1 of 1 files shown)


zeha (approved):
lgtm, thanks!

Issue #204, opened 2023-02-15
Disable new ext4 feature metadata_csum_seed feature for older Debian releases?


mika:
The annoying `metadata_csum` situation seems to reappear again with `metadata_csum_seed`, see https://bugs.debian.org/866603 + https://bugs.debian.org/1030846 for details. We _might_ need to check whether we're having e2fsprogs >=1.47.0-1 and then disable the feature when installing a release older than bookworm. :(

mika:
Same for `orphan_file`, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1031325#20

zeha:
> We might need to check whether we're having e2fsprogs >=1.47.0-1 and then disable the feature when installing a release older than bookworm.

Wouldn't always disabling the new feature when installing <= bookworm work, i.e. dont do the version check?

mika:
> Wouldn't always disabling the new feature when installing <= bookworm work, i.e. dont do the version check?

Yeah good idea! Didn't check though whether older e2fsprogs not being aware of this newer feature would complain when being invoked with ` -O ^metadata_csum_seed`, but if that works it would clearly simplify our live :) thx!

mika:
Yeah so if e2fsprogs is older than v1.43 it fails:

```
root@grml ~ # mkfs.ext4 -O '^metadata_csum_seed' /dev/sda1
mke2fs 1.42.13 (17-May-2015)
/dev/sda1 contains a ext4 file system
        created on Mon Feb 20 07:24:41 2023
Proceed anyway? (y,n) y
Invalid filesystem option set: ^metadata_csum_seed
```

zeha:
hmpf. that might be worthwhile feedback towards tytso.

PR #205, opened 2023-02-20
Disable ext4 metadata_csum_seed for Debian releases older than bookworm

mika:
As of e2fsprogs v1.43 and Linux kernel v4.4 the ext4 file system supports the new metadata_csum_seed feature (which allows the file system UUID to be modified without needing to update all of the file system metadata). Also see https://bugs.debian.org/1031325

Starting with e2fsprogs v1.47 this new metadata_csum_seed feature gets enabled by default. When installing an older Debian release this might cause problems because it's not fully supported yet, see e.g. the GRUB failure documented as #866603.

To keep the behavior identical to Debian, we do not enable the metadata_csum_seed feature for Debian releases before bookworm.  This is relevant for us, as grml-debootstrap might be running from a more recent Debian release (like a Debian testing/unstable based Grml live system).

We check for the e2fsprogs version as versions before 1.43 didn't support the metadata_csum_seed option yet.

Closes: https://bugs.debian.org/1031416
Closes: https://github.com/grml/grml-debootstrap/issues/204

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1311,6 +1311,24 @@ mkfs() {
     esac
   fi
 
+  # starting with e2fsprogs v1.47.0 mkfs.ext4 enables the metadata_csum_seed feature
+  # by default, which requires Linux kernel >=4.4, e2fsprogs >=1.43, according GRUB etc.
+  # Disable this feature for Debian releases older than bookworm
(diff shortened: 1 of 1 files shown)


zeha:
On first look this seems fine, gonna do a quick check tho.

zeha (approved):
Tested, seems good!

mika:
@zeha thanks for testing and quick response! :heart:

Issue #120, opened 2019-01-24
port to Multistrap or Similar for better Build Security


adrelanos:
Current situation and problem:

`build Debian base image using grml-debootstrap which uses debootstrap with mirror "packages.debian.org" but not "security.debian.org" because debootstrap can build only from one APT source` -> `might fetch vulnerable APT version like with DSA 4371` -> `vulnerable APT is used to apt-get update and dist-upgrade Debian base image using "packages.debian.org" and "security.debian.org"` -> `at this point the build might be compromised`

----

Proposed solution:

It would be better to use multistrap because multistrap can create Debian base images using multiple APT sources, "packages.debian.org" and "security.debian.org".

`build Debian base image using grml-debootstrap which uses multistrap with mirror "packages.debian.org"  and "security.debian.org"` -> `fetch security fixed packages only`

This doesn't only fetch security fixed APT but also never installs any already security fixed packages other than APT (think APT dependencies or anything other that is involved in processing code during installation such as libcurl perhaps).

----

Why? To be prepared for the next apt vulnerability.

* https://www.debian.org/security/2019/dsa-4371
* https://www.debian.org/security/2016/dsa-3733

adrelanos:
Quote https://wiki.debian.org/RISC-V (different context perhaps!) but it mentions `mmdebstrap` as alternative to `multistrap`.

> At the time of writing, mmdebstrap is still rather new, but it provides a largely debootstrap-compatible user interface and can be used as a drop-in replacement for debootstrap in many cases. multistrap already exists for many years, but it behaves quite a bit different from debootstrap and a chroot created with multistrap requires several manual adjustments that aren't necessary when using mmdebstrap. Unless there are specific reasons for using multistrap, the use of mmdebstrap is therefore recommended.

mika:
I'm not really buying the security argument, Debian provides point-releases (this time [9.7](https://packages.qa.debian.org/b/base-files/news/20190122T153207Z.html), very close to the apt 1.4.9 security release and meant exactly for that) and when bootstrapping stretch you get the fixed apt package already as intended, nor?

[...]

Issue #211, opened 2023-06-22
W: No zstd in /usr/bin:/sbin:/bin, using gzip


adrelanos:
During initramfs creation.

> W: No zstd in /usr/bin:/sbin:/bin, using gzip

Should `zstd` get installed by default?

mika:
Yeah, I think we can start shipping `zstd` by default. But given that it started to show up in 2015 within Debian, it would be interesting to know since which release it's available. (I'm not strictly saying we shouldn't modify the main `packages`, but I'm also reluctant about breaking backwards comparability. But maybe it's also good time to think about ways how to implement a better `packages` file handling, related to e.g. https://github.com/grml/grml-debootstrap/issues/44 + https://github.com/grml/grml-debootstrap/issues/98).

paulmenzel:
It’s in Debian 10 (buster). Are older releases supported by grml-debootstrap?

mika:
> It’s in Debian 10 (buster). Are older releases supported by grml-debootstrap?

Yes, see `releasetable.txt`:

```
.Current status
[width="40%",frame="topbot",options="header"]
|======================
|Release  |Status
|jessie   |works[1]
|stretch  |works[1]
|buster   |works[1]
|bullseye |works
|bookworm |works
|sid      |works[2]
|======================
```

(Reworking the package handling to make this per Debian release is on our wish list)

paulmenzel:
From [Debian BTS bug report #882244 (*zstd: add zstd package to jessie-backports*)](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=882244), I’d say it is *not* present in Debian 8 (jessie) proper and only in `jessie-backports`.

paulmenzel:
[ITP](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=806767)

PR #217, opened 2023-10-16
add mmdebstrap as alternative to (c)debootstrap

adrelanos:
https://github.com/grml/grml-debootstrap/issues/120

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -24,7 +24,7 @@ Package: grml-debootstrap
 Architecture: all
 Depends:
  debian-archive-keyring,
- debootstrap (>= 0.3.3.3) | cdebootstrap (>= 0.3.16),
+ debootstrap (>= 0.3.3.3) | cdebootstrap (>= 0.3.16) | mmdebstrap,
  e2fsprogs,
  fdisk | util-linux (<< 2.29.2-3~),
  gawk,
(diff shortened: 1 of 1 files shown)


mika:
Thanks, though I think we should also improve the code to actually fall back to usage of mmdebstrap if the default (debootstrap) isn't available. (I'd also like to get us to using mmdebstrap as the new default BTW :))

mika:
> Oh. It crossed my mind. But I was too shy to suggest mmdebstrap by default. :)
> 
> I thought with the goal of backwards compatibility as expressed in [#211 (comment)](https://github.com/grml/grml-debootstrap/issues/211#issuecomment-1606730725) I thought that might be hard.

Actually it shouldn't be _too_ hard to get this, even backwards compatible: one could run `DEBOOTSTRAP=mmdebstrap ...` for all the supported Debian releases and see whether it would break for anything. Even if it does break for older Debian releases, we could start enabling `mmdebstrap` for only more recent Debian releases? Or if necessary, also check that mmdebstrap is available with a minimum version and fall back to debootstrap else, if that should be relevant.

> #211 also looks easy to do except when attempting to keep backwards compatibility.

I just briefly checked, jessie doesn't seem to provide `zstd` yet, but it's available starting with `stretch`. (And runny wheezy isn't even easy any more with nowadays kernels due to `vsyscall=emulate` needs).  I'd tend to say, whoever needs to run something older than stretch should also use an older version of grml-debootstrap, and we can agree to consider stretch as the oldest support release from within grml-debootstrap then.

Issue #224, opened 2023-10-21
error not caught


adrelanos:
```
+ mount -t vfat /dev/mapper/loop0p1 /mnt/derivative-maker-grml-debootstrap.42265/boot/efi
mount: /mnt/derivative-maker-grml-debootstrap.42265/boot/efi: wrong fs type, bad option, bad superblock on /dev/mapper/loop0p1, missing codepage or helper program, or other error.
       dmesg(1) may have more information after failed mount system call.
+ chroot /mnt/derivative-maker-grml-debootstrap.42265 dpkg --list shim-signed
```
```
+ chroot /mnt/derivative-maker-grml-debootstrap.42265 grub-install --target=x86_64-efi --efi-directory=/boot/efi --uefi-secure-boot --removable /dev/loop0
Installing for x86_64-efi platform.
grub-install: error: /boot/efi doesn't look like an EFI partition.
+ chroot /mnt/derivative-maker-grml-debootstrap.42265 grub-install --target=i386-pc /dev/loop0
Installing for i386-pc platform.
Installation finished. No error reported.
```

Why this is happening is I need to investigate. Could be a localized issue. However, I can see REPORT_TRAP_ERR and FAIL_TRAP_ERR being `yes`. Yet, the error handler does not fire. The error is only caught later but not thorugh the `error_handler`.

```
+ umount /mnt/derivative-maker-grml-debootstrap.42265/boot/efi
umount: /mnt/derivative-maker-grml-debootstrap.42265/boot/efi: not mounted.
+ bailout 2 grub_install
+ cleanup
```

adrelanos:
I am pretty sure we need `set -e`.

I think REPORT_TRAP_ERR='yes' and FAIL_TRAP_ERR='yes' should be enabled by default anyhow. Even without any options needed. That's a good style. More reliable, secure. What do you think?

Otherwise conditionally enabling `set -e` when FAIL_TRAP_ERR='yes' is possible too but a less clean.

Could you advice please so I prepare a pull request?

adrelanos:
```
for i in format_efi_partition prepare_vm mkfs tunefs \
         mount_target mountpoint_to_blockdevice debootstrap_system \
         preparechroot execute_pre_scripts chrootscript execute_post_scripts \
         remove_configs umount_chroot grub_install umount_target fscktool ; do
    if stage "${i}" ; then
      if "$i" ; then
        stage "${i}" 'done' && rm -f "${STAGES}/${i}"
      else
[...]

PR #226, opened 2023-10-21
improve error handling

adrelanos:
* https://github.com/grml/grml-debootstrap/issues/224


mika:
We definitely need a reference to https://github.com/grml/grml-debootstrap/issues/224 here and further details in the commit message :) Otherwise I think it makes sense as-is, thanks

adrelanos:
Commit message improved just now.

mika:
Awesome, thanks!

Issue #229, opened 2023-10-23
`eend 0` does effectively nothing, so let's remove it


adrelanos:
```
eend() {
  local retval="${1:-0}"
  shift
  if [ "$retval" -gt 0 ]; then
    printf " %s-> Failed (rc=%s)%s\\n" "${BAD}" "${retval}" "${NORMAL}"
  fi
  return "$retval"
}
```

`eend 0` just results in `return 0` which doesn't do anything.

In preparation of https://github.com/grml/grml-debootstrap/issues/224 for code simplification it would be best to remove these.

A very few cases are a tiny bit more complex.

```
      if cp "$AUTHORIZED_KEYS_SOURCE" "$AUTHORIZED_KEYS_TARGET" ; then
        eend 0
      else
        eerror "Error: copying '$AUTHORIZED_KEYS_SOURCE' to '$AUTHORIZED_KEYS_TARGET' failed"
        eend 1
        bailout 1
      fi
```

Could rewrite to:

```
      if ! cp "$AUTHORIZED_KEYS_SOURCE" "$AUTHORIZED_KEYS_TARGET" ; then
        eerror "Error: copying '$AUTHORIZED_KEYS_SOURCE' to '$AUTHORIZED_KEYS_TARGET' failed"
        eend 1
        bailout 1
      fi
```

PR #230, opened 2023-10-23
code simplification

adrelanos:
Remove unnecessary use of `eend 0`.

fixes https://github.com/grml/grml-debootstrap/issues/229

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -289,7 +289,7 @@ cleanup() {
       fi
 
       if [ -n "$DIRECTORY" ] ; then
-        einfo "Not unmounting $MNTPOINT as you requested me to install into a directory of your own choice." ; eend 0
+        einfo "Not unmounting $MNTPOINT as you requested me to install into a directory of your own choice."
       else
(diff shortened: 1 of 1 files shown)


mika:
Ah, nice catch, thanks! That seems to be a leftover from c87dc4ae. 

nitpick: please slightly improve the commit message, the `code simplification` in the first line would show up as such in debian/changelog and doesn't tell much :)

So I'd propose something like:

```
code simplification: drop NOP usage of eend 0

Remove unnecessary use of `eend 0`, which does effectively nothing.
Leftover from git rev c87dc4ae where we moved from grml-etc-core's
lsb-functions functions to a simplified version of them.
    
Closes: https://github.com/grml/grml-debootstrap/issues/229
```

(Let me know if you'd prefer me to take care of it)

adrelanos:
Commit message updated.

I couldn't have invented that commit message myself. :)

mika:
> Commit message updated.

Great, thanks! Merged!
 
> I couldn't have invented that commit message myself. :)

:)

Issue #232, opened 2023-11-08
libpam-tmpdir breaks grml-debootstrap


adrelanos:
https://packages.debian.org/bookworm/libpam-tmpdir installed on the host operating system breaks grml-debootstrap.

```
+ chroot /mnt/derivative-maker-grml-debootstrap.10940 dpkg --list grub-pc
+ echo 'Notice: grub-pc package not present yet, installing it therefore.'
Notice: grub-pc package not present yet, installing it therefore.
+ DEBIAN_FRONTEND=noninteractive
+ chroot /mnt/derivative-maker-grml-debootstrap.10940 apt-get -y --no-install-recommends install -o Acquire::http::Proxy=http://127.0.0.1:3142 -o Acquire::https::Proxy=http://127.0.0.1:3142 -o Acquire::tor::Proxy=http://127.0.0.1:3142 -o APT::Update::Error-Mode=any -o Acquire::Languages=none -o Acquire::IndexTargets::deb::Contents-deb::DefaultEnabled=false -o Apt::Install-Recommends=false -o Acquire::Retries=5 -o Dpkg::Options::=--force-confnew grub-pc
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  dmsetup gettext-base grub-common grub-pc-bin grub2-common libbrotli1 libdevmapper1.02.1 libefiboot1 libefivar1 libfreetype6 libfuse2 libpng16-16 sensible-utils ucf
Suggested packages:
  multiboot-doc grub-emu mtools xorriso desktop-base console-setup fuse
Recommended packages:
  os-prober
The following NEW packages will be installed:
  dmsetup gettext-base grub-common grub-pc grub-pc-bin grub2-common libbrotli1 libdevmapper1.02.1 libefiboot1 libefivar1 libfreetype6 libfuse2 libpng16-16 sensible-utils
  ucf
0 upgraded, 15 newly installed, 0 to remove and 0 not upgraded.
Need to get 6068 kB of archives.
After this operation, 27.8 MB of additional disk space will be used.
...
Setting up grub-pc (2.06-13+deb12u1) ...
mktemp: failed to create file via template '/tmp/user/0/grub.XXXXXXXXXX': No such file or directory
dpkg: error processing package grub-pc (--configure):
 installed grub-pc package post-installation script subprocess returned error exit status 1
Processing triggers for libc-bin (2.36-9+deb12u3) ...
Errors were encountered while processing:
[...]

PR #231, opened 2023-10-25
improve error handling

adrelanos:
WIP

Not yet tested. Will test soon.

My commit messages are more explanatory for the reviewer than useful in the git history of the master branch.

Due the difficulty and extensive scope of these changes, I won't be able to split this into different logical sets of changes.  It was so extensive, I had to randomly start somewhere and adjust it step by step for the new error handler.

I would suggest to squash all of these commits into a single commit since functionality wise hopefully nothing changed, just the error handling was improved and the code simplified.

Fixes https://github.com/grml/grml-debootstrap/issues/224

PR #234, opened 2023-11-08
Fix compatibility with libpam-tmpdir.

adrelanos:
By creating folder the required temporary folder.

fixes https://github.com/grml/grml-debootstrap/issues/232

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1811,6 +1811,8 @@ preparechroot() {
   [ -n "$TUNE2FS" ]             && echo "TUNE2FS='$(sed "s,','\\\\'',g" <<<"${TUNE2FS}")'"                         >> "$CHROOT_VARIABLES"
   [ -n "$VMSIZE" ]              && echo "VMSIZE='$(sed "s,','\\\\'',g" <<<"${VMSIZE}")'"                           >> "$CHROOT_VARIABLES"
 
+  [ -n "$TMPDIR" ] && mkdir --parents "${MNTPOINT}/${TMPDIR}"
+
   cp $VERBOSE "${CONFFILES}"/chroot-script "${MNTPOINT}"/bin/chroot-script
   chmod 755 "${MNTPOINT}"/bin/chroot-script
   [ -d "$MNTPOINT"/etc/debootstrap/ ] || mkdir "$MNTPOINT"/etc/debootstrap/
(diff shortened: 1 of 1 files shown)


zeha:
I understand where you're coming from, but that seems to be the wrong fix.
Maybe we should reset TMPDIR when acting in the chroot instead.

adrelanos:
> I understand where you're coming from, but that seems to be the wrong fix.

Why? Some tool at some stage needs to create the directory. I was also wondering which tool would be the most appropriate. Maybe libpam-tmpdir should do it (or provider a helper tool to do that) but then I thought it won't be running at that time.

> Maybe we should reset TMPDIR when acting in the chroot instead.

Maybe. Then all invocations of `chroot` could be replaced with `$chroot` (or different variable name) which would prepend:

    env --unset=TEMP --unset=TEMPDIR --unset=TMP --unset=TMPDIR

What however should not be done is on the host operating system (at the beginning of grml-debootstrap):

    unset TEMP TEMPDIR TMP TMPDIR

Because that would disable libpam-tmpdir during all invocations of any tools running on the host operating system and not only inside the chroot.

On the other hand, wouldn't it be more secure to start using libpam-tmpdir as soon as possible even inside the chroot?

zeha:
> > I understand where you're coming from, but that seems to be the wrong fix.
> 
> Why? Some tool at some stage needs to create the directory. I was also wondering which tool would be the most appropriate. Maybe libpam-tmpdir should do it (or provider a helper tool to do that) but then I thought it won't be running at that time.

libpam-tmpdir on the *host* is a host thing. Its existence on the host shall not affect the chroot, which can end up on a completely different system / VM / ...

I know there are some other things that today leak from the host, but I'd like to get rid of them too - at least for `--vm` mode.

> > Maybe we should reset TMPDIR when acting in the chroot instead.
> 
> Maybe. Then all invocations of `chroot` could be replaced with `$chroot` (or different variable name) which would prepend:
> 
> ```
> env --unset=TEMP --unset=TEMPDIR --unset=TMP --unset=TMPDIR
> ```

I was thinking more along the lines of `env -i chroot ...`, as a holistic approach.

> On the other hand, wouldn't it be more secure to start using libpam-tmpdir as soon as possible even inside the chroot?

No, there is no concurrent access while the chroot is built. libpam-tmpdir does not add any value during this time.

Only after booting it might be useful, but this decision should be left to the user.

adrelanos:
Ok, very good. Lets define `$chroot`  being `env -i chroot` and use that?

I would attempt a PR but wait for it until https://github.com/grml/grml-debootstrap/pull/231 is merged (otherwise merge conflict).

zeha:
Instead of $chroot, we could have sth like  `clean_chroot() { env -i chroot "$@" }` (?)

adrelanos:
Works for me either way. `$chroot` has the advantage that it could be user customized through an environment variable.

> Instead of $chroot, we could have sth like `clean_chroot() { env -i chroot "$@" }` (?)

Since @mika liked the post, seems to agree with it (and I personally don't need the environment variable), I am happy to and will attempt to implement this at a later time when other PR is ready and merged to avoid merge conflicts.

PR #239, opened 2023-11-12
Skip EFI support check if --vm is given

zeha:
Hosts EFI support is unlikely to be related to the VMs EFI support, so this check often only adds noise.


zeha:

mika:
Good idea, thanks!

PR #241, opened 2023-11-13
Remove manual setup of /dev contents

zeha:
debootstrap already creates the minimally needed set of device nodes, and everything else will come from udev. Booting systems without udev is not supported in Debian TTBOMK, so we also do not need to try supporting that.

Here is the contents of /dev with this patch applied, for a bookworm install:

```
/dev:
total 4
drwxr-xr-x  4 root root  182 Nov 13 22:50 ./
drwxr-xr-x 17 root root 4096 Nov 13 22:51 ../
crw-rw-rw-  1 root root 5, 1 Nov 13 22:50 console
lrwxrwxrwx  1 root root   13 Nov 13 22:50 fd -> /proc/self/fd/
crw-rw-rw-  1 root root 1, 7 Nov 13 22:50 full
crw-rw-rw-  1 root root 1, 3 Nov 13 22:50 null
crw-rw-rw-  1 root root 5, 2 Nov 13 22:50 ptmx
drwxr-xr-x  2 root root    6 Nov 13 22:50 pts/
crw-rw-rw-  1 root root 1, 8 Nov 13 22:50 random
drwxr-xr-x  2 root root    6 Nov 13 22:50 shm/
lrwxrwxrwx  1 root root   15 Nov 13 22:50 stderr -> /proc/self/fd/2
lrwxrwxrwx  1 root root   15 Nov 13 22:50 stdin -> /proc/self/fd/0
lrwxrwxrwx  1 root root   15 Nov 13 22:50 stdout -> /proc/self/fd/1
crw-rw-rw-  1 root root 5, 0 Nov 13 22:50 tty
crw-rw-rw-  1 root root 1, 9 Nov 13 22:50 urandom
crw-rw-rw-  1 root root 1, 5 Nov 13 22:50 zero

/dev/pts:
total 0
drwxr-xr-x 2 root root   6 Nov 13 22:50 ./
drwxr-xr-x 4 root root 182 Nov 13 22:50 ../

/dev/shm:
total 0
drwxr-xr-x 2 root root   6 Nov 13 22:50 ./
drwxr-xr-x 4 root root 182 Nov 13 22:50 ../
```


zeha:

mika (commented):
This is *great*, I was only recently wondering whether this is still relevant at all, nice!

But now getting rid of this `devices.tar.gz` we could also remove the file from git, nor? :)

zeha:
> But now getting rid of this `devices.tar.gz` we could also remove the file from git, nor? :)

Good point. Done!

mika:
Thanks! :+1:

PR #242, opened 2023-11-13
Stop using sed when creating CHROOT_VARIABLES

zeha:
This breaks compatibility with running on a wheezy host, but bash in Debian >= jessie is already fixed.


zeha:

PR #243, opened 2023-11-14
Drop support for installing releases before jessie

zeha:
Whoever wants to install such an old release should please use an older grml-debootstrap.

This lets us drop old code, and going forward rely on new features.

Releases before jessie need vsyscall emulation. Jessie itself can only be installed by overriding both debootstrap's and apt's signing key checks. Stretch needs at least `--no-check-gpg` (or maybe another workaround), but for now it is installable.

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -730,26 +730,13 @@ grub_install() {
      done
      rm -f /boot/grub/device.map
   else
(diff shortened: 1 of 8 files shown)


zeha:

mika:
This totally makes sense, thanks!

PR #248, opened 2023-11-18
improve error handling 2

adrelanos:
https://github.com/grml/grml-debootstrap/pull/231

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -480,6 +480,10 @@ EOF
     local rootfs_mount_options=""
 
     if [ -z "${FILESYSTEM}" ] ; then
+      ## Debugging.
+      ls -la /dev/disk/by-uuid/"${TARGET_UUID}"
+      blkid /dev/disk/by-uuid/"${TARGET_UUID}"
+
       FILESYSTEM="$(blkid -o value -s TYPE /dev/disk/by-uuid/"${TARGET_UUID}")"
     fi
 
--- a/chroot-script
+++ b/chroot-script
@@ -806,7 +806,7 @@ trap signal_handler HUP INT QUIT TERM
      custom_scripts upgrade_system remove_apt_cache services \
      remove_chrootmirror; do
      if stage $i ; then
-       $i || exit 1
+       $i
        stage $i 'done'
      fi
   done
--- a/chroot-script
+++ b/chroot-script
@@ -501,8 +501,8 @@ EOF
   fi
 
 if [ -n "$EFI" ] ; then
-  # shellcheck disable=SC2086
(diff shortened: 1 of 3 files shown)


adrelanos:
`modprobe loop`

> modprobe: FATAL: Module loop not found in directory /lib/modules/6.2.0-1015-azure

Seems like an upstream issue as per https://github.com/microsoft/WSL/issues/10450

Previously without strict error handling, the CI simply didn't notice this issue. Solutions:

* A) One easy solution would be to ignore it using "`|| true`". This requires the least code. I'll do this for now to progress with testing on CI. Can be refined after I received feedback. Should I do this and add a comment with a link to that github issue?
* B) Maybe `modprobe loop` is no longer necessary and could be removed?
* C) Ignore this conditionally on CI only. Would require an "`if`", some environment variable or command line option such as `--ci`.

adrelanos:
> * A) One easy solution would be to ignore it using "`|| true`". This requires the least code. I'll do this for now to progress with testing on CI.

Done.

New issue. The following command has a non-zero exit code:

    blkid -o value -s TYPE /dev/disk/by-uuid/4f263a0c-b7b6-4998-b066-247fb3310d97

Any idea why that might be?

adrelanos:
This blkid related bug probably:

* existed earlier already,
* hasn't been spotted before,
* the new strict error handling in this PR simply made this bug visible,
* but this PR isn't the cause of this issue.

Can I ignore it for the purpose of this PR and create a follow-up issue instead? This is what I'll do for now so I can progress with further testing.

adrelanos:
blkid issue...

> Can I ignore it for the purpose of this PR and create a follow-up issue instead? This is what I'll do for now so I can progress with further testing.

Done.

Now, the CI is mostly passing except for the builds of Debian stretch are failing.

```
+ apt-get update
...
Err:5 http://security.debian.org/ stretch/updates/main amd64 Packages
  404  Not Found
...
W: The repository 'http://security.debian.org/ stretch/updates Release' does not have a Release file.
E: Failed to fetch http://security.debian.org/dists/stretch/updates/main/binary-amd64/Packages  404  Not Found
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

Not sure if this is an intermittent issue because Debian jessie does not have this issue?

Potential solutions:

* A) Dropping support for Debian stretch. This requires the least code. I am not a big fan of supporting legacy versions but you might have other requirements here.

Quote https://wiki.debian.org/DebianStretch

> Security updates have been discontinued as of 2020-07-06.

* B) Not enabling the security repository for Debian stretch builds.
* C) Adding `|| true` for invocations of `$APTUPDATE` (and similar) but only for Debian stretch builds

PR #250, opened 2023-11-18
ci: diff back-to-back built images

zeha:
Only known differences are accepted. Useful for finding new, unexpected differences.

The current list of accepted differences is a start; IMO we should shrink it.


adrelanos:
This is probably considered blocked by:

* https://github.com/grml/grml-debootstrap/issues/252

But I think it should not be considered blocked because of that. I think this PR would be good to be merged rather sooner than later to avoid merge conflicts. It seems to be a nice feature towards reproducible builds.

While the actual test is still broken on CI, it's still adding informative output to grml-debootstrap. I might even send a PR to improve that debug output. Also any scripts to diff two different VM images are nice for local use.

As long as https://github.com/grml/grml-debootstrap/issues/252 is pending, simply keep the test disabled on CI so CI does not break until this gets fixed in the future?

zeha:
> simply keep the test disabled on CI

we can just keep the PR open, no?

PR #255, opened 2023-11-22
improve error handling 3

adrelanos:
* https://github.com/grml/grml-debootstrap/pull/248
* https://github.com/grml/grml-debootstrap/pull/231

Fixes https://github.com/grml/grml-debootstrap/issues/224

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -104,8 +104,8 @@ chrootmirror() {
 
   # add security.debian.org:
   case "$RELEASE" in
-    unstable|sid) ;;  # no security pool available
-    jessie|stretch|buster)
+    unstable|sid|stretch) ;;  # no security pool available
+    jessie|buster)
       echo "Adding security.debian.org to sources.list."
       echo "deb http://security.debian.org ${RELEASE}/updates $COMPONENTS" >> /etc/apt/sources.list
       ;;
--- a/chroot-script
+++ b/chroot-script
@@ -619,7 +619,7 @@ efi_setup() {
 
   mkdir -p /boot/efi
   echo "Mounting $EFI on /boot/efi"
-  mount "$EFI" /boot/efi || return 1
+  mount "$EFI" /boot/efi
 
   # if efivarfs kernel module is loaded, but efivars isn't,
   # then we need to mount efivarfs for efibootmgr usage
@@ -629,7 +629,7 @@ efi_setup() {
   fi
 
   echo "Invoking efibootmgr"
-  efibootmgr || return 1
+  efibootmgr
 }
 
 # grub configuration/installation {{{
@@ -676,7 +676,7 @@ grub_install() {
     return 0
   fi
 
-  efi_setup || return 1
+  efi_setup
 
   if [ -n "$EFI" ] ; then
     GRUB_PACKAGE=grub-efi-amd64
--- a/chroot-script
+++ b/chroot-script
@@ -479,7 +479,7 @@ EOF
     local rootfs_mount_options=""
 
     if [ -z "${FILESYSTEM}" ] ; then
-      FILESYSTEM="$(blkid -o value -s TYPE /dev/disk/by-uuid/"${TARGET_UUID}")"
+      FILESYSTEM="$(blkid -o value -s TYPE /dev/disk/by-uuid/"${TARGET_UUID}")" || true
     fi
 
     case "${FILESYSTEM}" in
--- a/chroot-script
+++ b/chroot-script
@@ -804,7 +804,8 @@ trap signal_handler HUP INT QUIT TERM
      custom_scripts upgrade_system remove_apt_cache services \
      remove_chrootmirror; do
      if stage $i ; then
-       $i && stage $i 'done' || exit 1
+       $i || exit 1
+       stage $i 'done'
      fi
   done
   # always execute the finalize stage:
--- a/chroot-script
+++ b/chroot-script
@@ -803,9 +803,9 @@ trap signal_handler HUP INT QUIT TERM
      initrd grub_install passwords \
      custom_scripts upgrade_system remove_apt_cache services \
      remove_chrootmirror; do
-     if stage $i ; then
-       $i
-       stage $i 'done'
+     if stage "$i" ; then
+       "$i"
+       stage "$i" 'done'
      fi
   done
   # always execute the finalize stage:
--- a/chroot-script
+++ b/chroot-script
@@ -17,8 +17,10 @@ set -o pipefail
 trap "error_handler" ERR
 # }}}
 
+bash -n /etc/debootstrap/config
 # shellcheck source=config
 . /etc/debootstrap/config    || exit 1
+bash -n /etc/debootstrap/variables
 # shellcheck source=tests/shellcheck-stub-debootstrap-variables
 . /etc/debootstrap/variables || exit 1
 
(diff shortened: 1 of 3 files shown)


adrelanos:
Debian stretch build still failing as mentioned earlier in https://github.com/grml/grml-debootstrap/pull/248#issuecomment-1817766019.

But not an issue introduced by this PR. This PR here makes the issue only visible.

I compared with an arbitrary different recent PR. A PR not authored by me. Namely this PR:

* https://github.com/grml/grml-debootstrap/pull/248#issuecomment-1817766019

Here is the relevant excerpt from the other PR build log:

* https://github.com/grml/grml-debootstrap/actions/runs/6913689021/job/18810784414?pr=249

```
2023-11-18T11:48:47.2487870Z Err:7 http://security.debian.org stretch/updates/main amd64 Packages
2023-11-18T11:48:47.2488469Z   404  Not Found
2023-11-18T11:48:47.2529303Z Get:9 http://archive.debian.org/debian stretch/main Translation-en [5377 kB]
2023-11-18T11:48:47.2586326Z Ign:8 http://security.debian.org stretch/updates/main Translation-en
2023-11-18T11:48:48.6862318Z Fetched 5380 kB in 1s (2900 kB/s)
2023-11-18T11:48:49.1516508Z Reading package lists...
2023-11-18T11:48:49.1748396Z W: The repository 'http://security.debian.org stretch/updates Release' does not have a Release file.
2023-11-18T11:48:49.1750626Z E: Failed to fetch http://security.debian.org/dists/stretch/updates/main/binary-amd64/Packages  404  Not Found
2023-11-18T11:48:49.1752139Z E: Some index files failed to download. They have been ignored, or old ones used instead.
```

This issue seems parmement. And also not worth reporting to Debian as stretch doesn't have security support by Debian anymore anyhow.

Potential solutions:

* A) Dropping support for Debian stretch. This requires the least code. I am not a big fan of supporting legacy versions but you might have other requirements here.

Quote https://wiki.debian.org/DebianStretch

> Security updates have been discontinued as of 2020-07-06.

* B) Not enabling the security repository for Debian stretch builds.
* C) Adding `|| true` for invocations of `$APTUPDATE` (and similar) but only for Debian stretch builds. Not a fan of this.

I'll attempt to implement option B) for now until I receive other feedback.

adrelanos:
All tests are passing now except shellcheck but these seem to be unrelated to this PR.

I would appreciate some feedback on how to proceed with this PR.

grml-debootstrap
```diff
@@ -257,7 +257,9 @@ cleanup() {
 
   # Remove temporary mountpoint again
   if echo "$MNTPOINT" | grep -q '/mnt/debootstrap\.' ; then
-    rmdir "$MNTPOINT" 2>/dev/null
+    if test -d "$MNTPOINT" ; then
```
mika, line 260: please use `if [ -d ....` style here, as we do elsewhere in the code

chroot-script
```diff
@@ -805,9 +805,9 @@ trap signal_handler HUP INT QUIT TERM
      initrd grub_install passwords \
      custom_scripts upgrade_system remove_apt_cache services \
      remove_chrootmirror; do
-     if stage $i ; then
-       $i
-       stage $i 'done'
+     if stage "$i" ; then
```
mika, line 815: leaving the comment here, since on GitHub one sadly can't seem to comment on the commit message itself, AFAICS: the change itself LGTM, though please use something like "chroot-script: properly quote stage execution code" in the commit message

chroot-script
```diff
@@ -104,8 +104,8 @@ chrootmirror() {
 
   # add security.debian.org:
   case "$RELEASE" in
-    unstable|sid) ;;  # no security pool available
-    jessie|stretch|buster)
+    unstable|sid|stretch) ;;  # no security pool available
+    jessie|buster)
```
mika, line 108: are you sure this change makes sense? stretch is in between jessie and buster (see https://wiki.debian.org/DebianReleases)?

chroot-script
```diff
@@ -480,6 +480,10 @@ EOF
     local rootfs_mount_options=""
 
     if [ -z "${FILESYSTEM}" ] ; then
+      ## Debugging.
+      ls -la /dev/disk/by-uuid/"${TARGET_UUID}"
+      blkid /dev/disk/by-uuid/"${TARGET_UUID}"
```
mika: no reason to leave this commit in this PR, nor?

mika:
> All tests are passing now except shellcheck but these seem to be unrelated to this PR.

Ok great

> I would appreciate some feedback on how to proceed with this PR.

Yeah I think we can go forward with this, I left comments on some of the commits, and think we should get rid of the debugging related changes that shouldn't be part of the main grml-debootstrap? (No objections to enable debugging code when running with `--debug` option or alike though!)

chroot-script
adrelanos: Unfortunately, I am sure. The repository is defunct but not worth reporting upstream to Debian because that suite is end of life.

chroot-script
adrelanos: This means I should leave most commit messages as is but modify this specific commit message?

mika: I'd be more than happy to get nice commit messages! This one only clearly jumped into my eye :) So please feel free and invited to the reword your commit messages (and also squash logically related commits!), so they tell a nice story when going through them, but also have a (somewhat) nice state/story for debian/changelog (the first line of each commit message message shows up there by default when I release a new version, see `gbp dch`!)

chroot-script
adrelanos: In theory, I agree. It should also be affected. But in practice, it's not as I could see from the CI tests. I don't have an explanation why it is as is. I just noticed that oddly 1 Debian security repository for an EOL suite is broken.

adrelanos: Porting to `archive.debian.org` might fix this issue but lets consider that a separate issue that is not worth doing for an EOL suite? :)

adrelanos: Not that not enabling security repository for stretch didn't reduce the security or change installed packages.

* Previously: the stretch build would just show an error message that the stretch repository is unavailable, ignore it, and proceed.
* After this PR that enabled strict error checking: That resulted in a build failure so we at least know something is wrong. It only made this issue apparent. But the issue was there all along even if nobody noticed.This PR however didn't worsen anything. Well, could argue in theory if that stretch security repository ever comes back online then it's missing from these builds. If that is considered an issue then indeed code to make stretch use `archive.debian.org` might be required.

chroot-script
adrelanos: I was about to suggest, I was hoping, all of these commits could be squashed into 1. These changes aren't really interesting for users. Just "improved error handling".

chroot-script
adrelanos: Ok. Removed after force push. Done.

chroot-script
adrelanos: I rewrote this commit message, grouped together and also improved other commit messages.

adrelanos:
I've hopefully addressed all your feedback. Please kindly let me know if there's other blockers to get this merged.

There's a new related issue which is triggered (but not caused) by this PR for which I created a separate issue:

* https://github.com/grml/grml-debootstrap/issues/256

mika:
Sorry for the delay, finally found some spare minutes to take care of this. Thanks for all your work around this, @adrelanos ! :+1:

PR #259, opened 2023-12-12
consistently use ESP label for the EFI system partition

adrelanos:
change from EFI to ESP for arm64 builds

Use of ESP is more popular.

* https://en.wikipedia.org/wiki/EFI_system_partition
* https://wiki.archlinux.org/title/EFI_system_partition
* https://www.diskpart.com/articles/efi-system-partition-4348.html

If we'd use the long label it would be "EFI system partition" but since we use the short label it should be ESP.

Why bother? Because then later `PARTLABEL` can be used in `/etc/fstab`.

https://wiki.archlinux.org/title/fstab#GPT_partition_labels

Any label with spaces is more error-prone.

I haven't found any convention for partition labels but consistency makes sense.

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1517,7 +1517,7 @@ prepare_vm() {
     if [ "$ARCH" = 'arm64' ]; then
       einfo "Setting up GPT partitions for arm64"
       parted -s "${TARGET}" 'mklabel gpt'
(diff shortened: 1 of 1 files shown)


mika:
Wasn't aware of this, nice catch, thanks! :+1:

PR #262, opened 2023-12-22
improved error handling, removed RC=$? noops

adrelanos:


grml-debootstrap
```diff
@@ -906,10 +906,12 @@ while IFS= read -r i; do
 done < "$TMPFILE"
 
 ERRORFILE=$(mktemp)
+
+local RC=0
 # shellcheck disable=SC2086
 yes | mdadm --create "${TARGET}" --level="${RAIDLEVEL}" \
-      --raid-devices="${NUM_PARTITIONS}" ${SELECTED_PARTITIONS} >/dev/null 2>$ERRORFILE
-RC=$?
+      --raid-devices="${NUM_PARTITIONS}" ${SELECTED_PARTITIONS} >/dev/null 2>$ERRORFILE || RC=$?
```
adrelanos, line 913: The github diff looks weird. But if you look at the full file it looks simple and correct.

```
local RC=0
# shellcheck disable=SC2086
yes | mdadm --create "${TARGET}" --level="${RAIDLEVEL}" \
      --raid-devices="${NUM_PARTITIONS}" ${SELECTED_PARTITIONS} >/dev/null 2>$ERRORFILE || RC=$?

if [ "$RC" = 0 ] ; then
   dialog --title "$PN" --msgbox \
   "Creating $TARGET was successful." 0 0
   rm -f "$TMPFILE" "$ERRORFILE"
else
```

Could have used:

    if yes | mdadm ... ; then

but given the already long and complex mdadm command, I opted for a simpler fix. That is, on top of the command `local RC=0` and at the end of the command `|| RC=$?`.

PR #265, opened 2023-12-23
ci: enable test-build on pushes to branches and periodic

zeha:
Enables test builds on pushes, to all branches. Assumes people usually work on their own fork. They can enable/disable GitHub Actions on their forks as needed.

Also enables a periodic test build (on Tuesdays), so we know when Debian testing breaks or similar.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,6 +1,10 @@
 name: test-build
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


mika:
Great idea, thanks! :heart:

mika (approved):

zeha:
Was merged manually

PR #266, opened 2023-12-23
use apt-get with --error-on=any option

adrelanos:
for better error handling

Otherwise if 1 repository fails (such as the security repository) apt would by default silently ignore this can continue the build. Therefore for consistent builds and towards reproducible builds, this option is required.

I am using this in all my build scripts for years, though injected through DPKG_OPTIONS but I think the case is strong enough to hardcode this as default.

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -42,7 +42,7 @@ if [ -x /usr/bin/aptitude ] ; then
    fi
 else
    APTINSTALL="apt-get -y --no-install-recommends install $DPKG_OPTIONS"
-   APTUPDATE="apt-get update $DPKG_OPTIONS"
+   APTUPDATE="apt-get --error-on=any update $DPKG_OPTIONS"
    APTUPGRADE="apt-get -y upgrade $DPKG_OPTIONS"
 fi
 


adrelanos:
A simple and non-controversial change?

mika:
The problem I have with this change is, that this feature seems to be available as of apt v2.1.16, which is available "only" as of bullseye (v11) and newer Debian releases, while we still supported Debian releases back until jessie (v8).

The `22 successful and 18 failing checks` also indicates that this might be an actual problem, as we currently usually have "only" 10 failing jobs (known as https://github.com/grml/grml-debootstrap/issues/278).

So I'm afraid while the change looks simple and non-controversial, it's not as simple as that. :-/ (Furthermore in line 182 of `chroot-script` we have another invocation of `apt-get update` that we should consider updating as well, possibly also the ones in `docker/Dockerfile`, `packer/` and `tests/` even.)

zeha:
I guess feature detection could be added, then ok?

mika:
> I guess feature detection could be added, then ok?

ACK! :)

mika:
> > I guess feature detection could be added, then ok?
> 
> ACK! :)

Uff, the option isn't mentioned in `apt-get`'s help output and also isn't reported with a separate exit code either (only the 100 one which is used for everything™):

```
root@e10866c823d0:/# cat /etc/debian_version 
10.13
root@e10866c823d0:/# apt-get --error-on=any 
E: Command line option --error-on=any is not understood in combination with the other options
root@e10866c823d0:/# apt-get --error-on=any update
E: Command line option --error-on=any is not understood in combination with the other options
root@e10866c823d0:/# echo $?
100
root@e10866c823d0:/# apt-get --help
apt 1.8.2.3 (amd64)
Usage: apt-get [options] command
       apt-get [options] install|remove pkg1 [pkg2 ...]
       apt-get [options] source pkg1 [pkg2 ...]

apt-get is a command line interface for retrieval of packages
and information about them from authenticated sources and
for installation, upgrade and removal of packages together
with their dependencies.

Most used commands:
  update - Retrieve new lists of packages
  upgrade - Perform an upgrade
  install - Install new packages (pkg is libc6 not libc6.deb)
  reinstall - Reinstall packages (pkg is libc6 not libc6.deb)
  remove - Remove packages
  purge - Remove packages and config files
  autoremove - Remove automatically all unused packages
  dist-upgrade - Distribution upgrade, see apt-get(8)
  dselect-upgrade - Follow dselect selections
  build-dep - Configure build-dependencies for source packages
  clean - Erase downloaded archive files
  autoclean - Erase old downloaded archive files
  check - Verify that there are no broken dependencies
  source - Download source archives
  download - Download the binary package into the current directory
  changelog - Download and display the changelog for the given package

See apt-get(8) for more information about the available commands.
Configuration options and syntax is detailed in apt.conf(5).
Information about how to configure sources can be found in sources.list(5).
Package and version choices can be expressed via apt_preferences(5).
Security details are available in apt-secure(8).
                                        This APT has Super Cow Powers.
```

Any ideas how to properly detect `--error-on=...` support? :thinking:

adrelanos:
Would probably need to set or not set the option based on the Debian version.

zeha:
Rebased this to get current pipelines

zeha:
As expected at least buster fails.

mika (changes requested):
We need to at least find a workaround for buster to be able to accept this change, at least as long as we support Debian/buster also.

mika:
Now that we dropped support for buster and only target bullseye and newer, we could maybe take care of this now? @adrelanos

PR #267, opened 2023-12-23
implement clean_chroot,

adrelanos:
avoid host environment variables such as TMP to leak into the chroot

fixes https://github.com/grml/grml-debootstrap/issues/232

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1766,7 +1766,7 @@ clean_chroot() {
   local chroot_command=("${@:2}")
 
   # Run chroot, then env -i with the specified environment variables inside the chroot
-  /usr/sbin/chroot "$chroot_dir" /usr/bin/env -i "${env_vars[@]}" "${chroot_command[@]}"
+  chroot "$chroot_dir" /usr/bin/env -i "${env_vars[@]}" "${chroot_command[@]}"
 }
 # }}}
 
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -13,8 +13,7 @@ error_handler() {
    last_bash_command="$BASH_COMMAND"
    echo "Unexpected non-zero exit code $last_exit_code in ${BASH_SOURCE[*]} at line ${BASH_LINENO[*]} detected!
 last bash command: $last_bash_command"
-   ## Check if "bailout" function is available.
-   ## This is not the case in chroot-script.
+   ## Check if "bailout" function is already available.
    if command -v bailout >/dev/null 2>&1; then
       bailout 1
    else
@@ -26,7 +25,6 @@ set -e
 set -E
 set -o pipefail
 trap "error_handler" ERR
-export -f "error_handler"
 # }}}
 
 # variables {{{
@@ -74,11 +72,6 @@ MNTPOINT="/mnt/debootstrap.$$"
 [ -n "$UPGRADE_SYSTEM" ] || UPGRADE_SYSTEM='yes'
 [ -n "$VMSIZE" ] || VMSIZE="2G"
 [ -n "$GRUB_INSTALL" ] || GRUB_INSTALL='yes'
-
-# inside the chroot system locales might not be available, so use minimum:
-export LANG=C
-export LC_ALL=C
-export LANGUAGE=C
 
 # make sure interactive mode is only executed when
 # using an empty configuration file or option --interactive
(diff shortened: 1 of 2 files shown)


adrelanos:
Some optional explanation why I made certain changes the way I made them.

If we use `env -i` then we can no longer export shell functions. So `export -f "error_handler"` had to be removed.

`clean_chroot` function has become more complex than we might have hoped for.

Even `PATH` needs to be set. Otherwise `clean_chroot "$MNTPOINT" grub-install` would fail because `grub-install` is in `/usr/sbin/grub-install` in the chroot.

A potential alternative that I personally would use is `sudo` as that already has sane environment scrubbing but I wasn't sure you would want a dependency on `sudo`. Also maybe in the future you would want to run most of grml-deboostrap as a unprivileged user and only use a privilege escalation tool (such as sudo) when needed and/or run commands unprivileged inside the chroot.

`http_proxy` has to be passed otherwise apt-cacher-ng would be broken by this commit. While at it, I completed it and added `https_proxy`, and `ALL_PROXY` there too for completeness sake.

Which environment variables are passed into the chroot is currently hardcoded. I didn't think it is important to be able to configure that. If somebody needs to configure that, I guess they can create a feature request and/or send a pull request.

The xtrace (`bash -x`) could be minimized if the code to set the environment variables was just run and by using global variables. That however would come at the cost at slightly more code complexity.

I was also wondering if it was better to use a similar mechanism to the one you're using for `CHROOT_VARIABLES`, but that would not work because only the `chroot-script` reads those. But you're not only using that but also other calls from grml-debootstrap to `chroot` (now `clean_chroot`) so the environment variables need to be set at the grml-debootstrap level.

adrelanos:
Ping?

grml-debootstrap
```diff
@@ -1737,6 +1730,46 @@ debootstrap_system() {
...
+  done
+
+  # First argument is the chroot directory.
+  local chroot_dir="$1"
+
+  # Remaining arguments are for the command to be executed in the chroot environment.
+  local chroot_command=("${@:2}")
+
+  # Run chroot, then env -i with the specified environment variables inside the chroot
+  /usr/sbin/chroot "$chroot_dir" /usr/bin/env -i "${env_vars[@]}" "${chroot_command[@]}"
```
mika, line 1792: Hm, why the explicit call to `/usr/sbin/chroot` instead of relying on $PATH?

adrelanos: Done.

adrelanos:
Good now?

adrelanos:
Ping?

grml-debootstrap
```diff
@@ -1766,7 +1766,7 @@ clean_chroot() {
   local chroot_command=("${@:2}")
 
   # Run chroot, then env -i with the specified environment variables inside the chroot
-  /usr/sbin/chroot "$chroot_dir" /usr/bin/env -i "${env_vars[@]}" "${chroot_command[@]}"
+  chroot "$chroot_dir" /usr/bin/env -i "${env_vars[@]}" "${chroot_command[@]}"
```
mika, line 1792: I think we even don't need to hardcode the `/usr/bin/env` full path?

grml-debootstrap
```diff
@@ -13,7 +13,14 @@ error_handler() {
...
 last bash command: $last_bash_command"
-   ## Check if "bailout" function is already available.
+   if [ -r "$MNTPOINT/debootstrap/debootstrap.log" ] && \
+          [ -s "$MNTPOINT/debootstrap/debootstrap.log" ] ; then
+       einfo "Presenting last ten lines of debootstrap.log:"
+       tail -10 "${MNTPOINT}"/debootstrap/debootstrap.log
+       einfo "End of debootstrap.log"
+   fi
+   ## Check if "bailout" function is available.
+   ## This is not the case in chroot-script.
```
mika, line 23: hm, I don't understand why this change ends up in this clean_chroot feature/commit, is the github web UI showing something unexpected? o_

grml-debootstrap
```diff
@@ -26,7 +25,6 @@ set -e
 set -E
 set -o pipefail
 trap "error_handler" ERR
-export -f "error_handler"
```
mika, line 35: is removing the export of the error_handler from `grml-debootstrap` and addition of it to chroot script related to the environment cleanup, right?

mika: Ah, you mentioned this in a comment as I just noticed:

> If we use env -i then we can no longer export shell functions. So export -f "error_handler" had to be removed.

We should squash all the relevant commits into one single commit and provide that useful information you collected into the commit message :)

Do you want me to give this a try or would you prefer to try on your own?

grml-debootstrap
adrelanos: True. Changed to `env`.

grml-debootstrap
adrelanos: I am not sure where you saw that. But GitHub says `Outdated`". I cannot see any change related to / near `## Check if "bailout" function is available.` in a local git diff.

grml-debootstrap
adrelanos: > Do you want me to give this a try

Yes, please.

adrelanos:
Ping?

mika:
> Ping?

Sorry4delay, finally found some time to look into this, see https://github.com/grml/grml-debootstrap/pull/280

PR #270, opened 2024-01-27
enable `discard` mount option by default in `/etc/fstab`

adrelanos:
This is for easier `trim` support in VMs to release the free space on the host operating system after deleting files inside the VM image.

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -492,7 +492,7 @@ EOF
     case "${FILESYSTEM}" in
       # errors=remount-ro is supported only by a few file systems
       ext*|exfat|fat|jfs|nilfs2|vfat)
-        rootfs_mount_options=",errors=remount-ro"
+        rootfs_mount_options="discard,errors=remount-ro"
         ;;
     esac
 


adrelanos:
Do you know if this causes any issues for any of the file systems? I am only using `ext4`.

>      ext*|exfat|fat|jfs|nilfs2|vfat)

Ideally file systems not supporting it would simply ignore it.

Otherwise this would need a different implementation doing this only where known to be functional, for `ext4` with others happy to send patches to opt this in for other file systems.

mika:
AFAICS all those file systems seem to support `discard`, so I'd give this a try.

@zeha @jkirk what do you think?

jkirk:
I am not so sure about changing the default mount options.

`ext4(5)` states:

>        discard/nodiscard
>               Controls whether ext4 should issue discard/TRIM commands to the underlying block device when blocks are freed. 
>               This is useful for SSD devices and sparse/thinly-provisioned LUNs,
>               but it is off by default until sufficient testing has been done.

`xfs(5)` states:

>        discard|nodiscard
>               Enable/disable the issuing of commands to let the block
>               device reclaim space freed by the filesystem.  This is
>               useful for SSD devices, thinly provisioned LUNs and
>               virtual machine images, but may have a performance impact.
> 
>               Note: It is currently recommended that you use the fstrim
>               application to discard unused blocks rather than the
>               discard mount option because the performance impact of
>               this option is quite severe.  For this reason, nodiscard
>               is the default.

I could not find a man page for `mount.exfat`.
`mount.exfat(8)` from exfat-fuse does not mention `discard` and mount(8) does not mention `exfat` at all.

`mount(8)` does not mention `discard` for `jfs`.

`mount.nilfs2(8)` mentions `discard`, but AFAICS `nilfs-tools` is needed for `mkfs.nilfs2`; shouldn't it be installed via [install_fs_tools()](https://github.com/grml/grml-debootstrap/blob/38c94d1903c6bede00b7ffce3dc3c26b93493596/chroot-script#L538)?

And finally `fstrim(8)`

> Running fstrim frequently, or even using mount -o discard, might
>        negatively affect the lifetime of poor-quality SSD devices. For
>        most desktop and server systems a sufficient trimming frequency
>        is once a week. Note that not all devices support a queued trim,
>        so each trim command incurs a performance penalty on whatever
>        else might be trying to use the disk at the time.

If at all, I'd vote to make `discard` optional.

adrelanos:
Yes. Thank you for your consideration!

Closing.

PR #277, opened 2024-06-07
GitHub Actions: Add workflow_dispatch trigger

jkirk:
To manually run a workflow,  the workflow must be configured to run on the workflow_dispatch event.

We want that all our GitHub Actions can be triggered manually.

See:

* https://docs.github.com/en/actions/using-workflows/manually-running-a-workflowk
* https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -3,6 +3,7 @@
 
 name: pr-review
 on:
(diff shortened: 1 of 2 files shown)


mika (approved):
Thx!

PR #279, opened 2024-08-12
Tag installed policy-rc.d file

zeha:
Useful in case it is not cleaned up and later on found on an installed system.


mika:
Great idea, thanks! :+1:

PR #280, opened 2024-08-14
Implement clean_chroot to avoid host env variables like TMP to leak into the chroot

mika:
Some implementation notes:

If we use `env -i`, then we can no longer export shell functions. So export -f "error_handler" had to be removed.

`PATH` needs to be set, otherwise `clean_chroot "$MNTPOINT" grub-install` would fail, because grub-install is in /usr/sbin/grub-install in the chroot.

http_proxy has to be passed otherwise apt-cacher-ng would be broken by this commit.  While at it, I completed it and added https_proxy, and ALL_PROXY there too for completeness sake.

Which environment variables are passed into the chroot is currently hardcoded.

FTR, I was also wondering if it was better to use a similar mechanism to the one you're using for CHROOT_VARIABLES, but that would not work because only the chroot-script reads those. But we're not only using that but also other calls from grml-debootstrap to chroot (now clean_chroot), so the environment variables need to be set at the grml-debootstrap level.

Closes: grml/grml-debootstrap#232

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -11,6 +11,14 @@
 # shellcheck disable=SC2317  # shellcheck has trouble understanding the code flow in this file
 
 # error_handler {{{
+error_handler() {
+   last_exit_code="$?"
+   last_bash_command="$BASH_COMMAND"
+   echo "Unexpected non-zero exit code $last_exit_code in ${BASH_SOURCE[*]} at line ${BASH_LINENO[*]} detected!
+last bash command: $last_bash_command"
+   exit 1
+}
+
 set -e
 set -E
 set -o pipefail
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -19,8 +19,7 @@ last bash command: $last_bash_command"
        tail -10 "${MNTPOINT}"/debootstrap/debootstrap.log
        einfo "End of debootstrap.log"
    fi
-   ## Check if "bailout" function is available.
-   ## This is not the case in chroot-script.
+   ## Check if "bailout" function is already available.
    if command -v bailout >/dev/null 2>&1; then
       bailout 1
    else
@@ -32,7 +31,6 @@ set -e
 set -E
 set -o pipefail
 trap "error_handler" ERR
-export -f "error_handler"
 # }}}
 
 # variables {{{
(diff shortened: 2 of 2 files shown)


mika:
This is the result of squashing the commits from https://github.com/grml/grml-debootstrap/pull/267 and extending the commit message accordingly.

What I'm not entirely sure about yet is the usage of calls like:

```
clean_chroot "$MNTPOINT" DEBIAN_FRONTEND=$DEBIAN_FRONTEND apt-get [...]
```

@adrelanos are you sure this behaves as expected? :thinking: 
Would it make sense to support `DEBIAN_FRONTEND` via the `additional_vars` you implemented in `clean_chroot()`? :thinking:

adrelanos:
> This is the result of squashing the commits from #267 and extending the commit message accordingly.

Thank you!

> What I'm not entirely sure about yet is the usage of calls like:
> 
> ```
> clean_chroot "$MNTPOINT" DEBIAN_FRONTEND=$DEBIAN_FRONTEND apt-get [...]
> ```
> 
> @adrelanos are you sure this behaves as expected? 🤔 

I've built several releases using that line. It is functional because we are using `/usr/bin/env -i` and are therefore allowed to extend that.

    chroot bookworm /usr/bin/env -i testvar=test env | grep testvar

> testvar=test

> Would it make sense to support `DEBIAN_FRONTEND` via the `additional_vars` you implemented in `clean_chroot()`? 🤔

It would make sense but I did not include it to keep this PR smaller. It wasn't strictly required, I thought. And it made the diff look simpler. I could be done in a follow-up.

mika:
> > What I'm not entirely sure about yet is the usage of calls like:
> > ```
> > clean_chroot "$MNTPOINT" DEBIAN_FRONTEND=$DEBIAN_FRONTEND apt-get [...]
> > ```
> > @adrelanos are you sure this behaves as expected? 🤔
> 
> I've built several releases using that line. It is functional because we are using `/usr/bin/env -i` and are therefore allowed to extend that.

Ahhhh right, now I see the magic in `Capture additional environment variables passed as arguments` that makes that possible, thanks :)

> ```
> chroot bookworm /usr/bin/env -i testvar=test env | grep testvar
> ```
> 
> > testvar=test
> 
> > Would it make sense to support `DEBIAN_FRONTEND` via the `additional_vars` you implemented in `clean_chroot()`? 🤔
> 
> It would make sense but I did not include it to keep this PR smaller. It wasn't strictly required, I thought. And it made the diff look simpler. I could be done in a follow-up.

Ok :)

adrelanos:
Mergable?

mika:
> Mergable?

Had to rebase and fix conflicts, taking care via https://github.com/grml/grml-debootstrap/pull/new/mika/clean_chroot

Issue #278, opened 2024-06-08
Github Actions - Workflow test-build : 10 errors and 39 warnings


jkirk:
We get 10 errors and 39 warnings in the Github Workflow [test-build](https://github.com/grml/grml-debootstrap/actions/runs/9418427334):

The 10 errors come from:

```
test-debian (trixie, bullseye, mmdebstrap)
Process completed with exit code 1.
test-debian (trixie, stretch)
Process completed with exit code 1.
test-debian (trixie, buster)
Process completed with exit code 1.
test-debian (unstable, stretch)
Process completed with exit code 1.
test-debian (trixie, bullseye)
Process completed with exit code 1.
test-debian (unstable, bullseye, mmdebstrap)
Process completed with exit code 1.
test-debian (unstable, bullseye)
Process completed with exit code 1.
test-debian (trixie, buster, mmdebstrap)
Process completed with exit code 1.
test-debian (unstable, buster, mmdebstrap)
Process completed with exit code 1.
test-debian (unstable, buster) 
```

I took a closer look into the logs of `test-debian (unstable, buster)` and saw the following:
```
Run ./tests/build-vm-and-test.sh test
+ '[' '!' -d ./tests ']'
+ '[' test == setup ']'
+ RELEASE=buster
+ TARGET=qemu.img
+ DEBOOTSTRAP=
[...]
+ qemu-system-x86_64 -hda /home/runner/work/grml-debootstrap/grml-debootstrap/qemu.img -m 2048 -display none -vnc :0 -virtfs local,path=/tmp/tmp.viVk190zWP,mount_tag=host0,security_model=none,id=host0 -serial pty
+ echo 'No serial console from Qemu found yet [29 retries left]'
No serial console from Qemu found yet [29 retries left]
[...]
+ /home/runner/work/grml-debootstrap/grml-debootstrap/tests/serial-console-connection --tries 180 --port /dev/pts/0 --hostname buster --poweroff 'mount -t 9p -o trans=virtio,version=9p2000.L,rw host0 /mnt && cd /mnt && ./testrunner'
Login failure (try 0): Timeout exceeded.
[...]
Login failure (try 101): Timeout exceeded.
<pexpect.fdpexpect.fdspawn object at 0x7f7222758e50>
searcher: searcher_re:
    0: re.compile(b'buster login:')
Logging into /dev/pts/0 via serial console [try 0]
Waiting for login prompt...
Logging into /dev/pts/0 via serial console [try 1]
[...]

PR #284, opened 2024-08-17
tests: capture screenshot on failure

zeha:
@mika wrote most of this, I just rebased it on top of #283 

- Capture VM screenshot when tests fail.
- Add instructions for running the tests locally.
- Ensure curl is installed.
- Update default hostname in helper script.

Debugging aid for #278.


zeha:
Example result:

screenshot

mika (approved):
Thanks, LGTM!

PR #287, opened 2024-08-17
tests: set msize=512000 option for 9pfs

zeha:
From https://salsa.debian.org/ci-team/autopkgtest/-/merge_requests/418

We do not transfer large file, but still.


mika (approved):
Nice one, thx!

zeha:
Rebased so we can actually merge. Also gives us a new test run :)

Issue #290, opened 2024-10-20
Use mmdebstrap as default


zeha:
Necessary for grml-small. And generally a good idea.

zeha:
@mika wrote:

> I'm actually wondering whether we should drop cdebootstrap support, I never ever use that, I think it's worth dropping this?

Agree, we should probably drop cdebootstrap support. I think it's dead in Debian, too.

We should also considering dropping debootstrap support. Doing that will allow us being clear about the tool that will run, remove a config var, simplify the test matrix, etc. And IMO mmdebstrap is strictly the better tool - if available.

Doing that might mean killing support for running on `stretch`. Or we figure out what to do about `--no-check-gpg` for stretch's mmdebstrap.

adrelanos:
Does debootstrap have better cross architecture build support?

Or qemu-debootstrap?

I think it's worth keeping support for changing debootstrap tool for 
future-proofness.

adrelanos:
Switching the default however seems nice.

zeha:
> Does debootstrap have better cross architecture build support?
> Or qemu-debootstrap?

Even if they do, grml-debootstrap has IIRC no cross support.

adrelanos:
grml-debootstrap has cross build support. Building ARM64 (on AMD64) was implemented here:

* https://github.com/grml/grml-debootstrap/issues/169
* https://github.com/grml/grml-debootstrap/pull/191

zeha:
> grml-debootstrap has cross build support. Building ARM64 (on AMD64) was implemented here:
> 
> * [ARM support #169](https://github.com/grml/grml-debootstrap/issues/169)
> * [Initial arm64 support #191](https://github.com/grml/grml-debootstrap/pull/191)

Right. Then we should also run CI for that.

Issue #209, opened 2023-04-10
zoom not starting


suntong:
Following up on #208, in which my `/etc/debootstrap/packages` were listed.

I then installed `zoom` from https://zoom.us/client/latest/zoom_amd64.deb, but when start `zoom`, nothing happens:

```sh
$ time zoom | wc
      0       0       0

real    0m0.487s
user    0m0.208s
sys     0m0.242s
```

suntong:
Hmm... maybe missing some important QT libs.

```sh
$ cat ~/.zoom/logs/zoom_stdout_stderr.log
. . .
Zoom package arch is 64bit, runing OS arch is x86_64, snap package 0
zoom started.
/lfs/workv/opt/zoom/zoom : symbol lookup error: /lfs/workv/opt/zoom/QtQuick/Controls.2/libqtquickcontrols2plugin.so: undefined symbol: _ZN13QQmlDirParserC1Ev, version Qt_5_PRIVATE_API
Something went wrong while running zoom, exit code is 127.
ZoomLauncher exit.
```

I'll close it and look for solutions elsewhere.

suntong:
Hmm... maybe missing some important QT libs.

```sh
$ cat ~/.zoom/logs/zoom_stdout_stderr.log
. . .
Zoom package arch is 64bit, runing OS arch is x86_64, snap package 0
zoom started.
/lfs/workv/opt/zoom/zoom : symbol lookup error: /lfs/workv/opt/zoom/QtQuick/Controls.2/libqtquickcontrols2plugin.so: undefined symbol: _ZN13QQmlDirParserC1Ev, version Qt_5_PRIVATE_API
Something went wrong while running zoom, exit code is 127.
ZoomLauncher exit.
```

I'll close it and look for solutions elsewhere.

PR #291, opened 2024-10-20
Switch default debootstrap tool to mmdebstrap

zeha:
mmdebstrap is the modern, fast tool. grml-small comes with mmdebstrap but not debootstrap, so it's really necessary there.

Closes: #290


mika (commented):
I think we should also adjust `README.md`:

```
-This tool is a wrapper suite around debootstrap and cdebootstrap
+This tool is a wrapper suite around debootstrap and mmdebstrap
```

and also adapt `grml-debootstrap.8.txt` accordingly?

zeha:
> I think we should also adjust `README.md`:
> 
> ```
> -This tool is a wrapper suite around debootstrap and cdebootstrap
> +This tool is a wrapper suite around debootstrap and mmdebstrap
> ```
> 
> and also adapt `grml-debootstrap.8.txt` accordingly?

Indeed, done.

I have left the parts mentioning just debootstrap alone for now, we can continue the discussion in #209.

PR #295, opened 2024-11-22
Implement clean_chroot to avoid host env variables like TMP to leak i…

mika:
…nto the chroot

Some implementation notes:

If we use `env -i`, then we can no longer export shell functions. So export -f "error_handler" had to be removed.

`PATH` needs to be set, otherwise `clean_chroot "$MNTPOINT" grub-install` would fail, because grub-install is in /usr/sbin/grub-install in the chroot.

http_proxy has to be passed otherwise apt-cacher-ng would be broken by this commit.  While at it, I completed it and added https_proxy, and ALL_PROXY there too for completeness sake.

Which environment variables are passed into the chroot is currently hardcoded.

FTR, I was also wondering if it was better to use a similar mechanism to the one you're using for CHROOT_VARIABLES, but that would not work because only the chroot-script reads those. But we're not only using that but also other calls from grml-debootstrap to chroot (now clean_chroot), so the environment variables need to be set at the grml-debootstrap level.

Closes: grml/grml-debootstrap#232

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -11,6 +11,14 @@
 # shellcheck disable=SC2317  # shellcheck has trouble understanding the code flow in this file
 
 # error_handler {{{
+error_handler() {
+   last_exit_code="$?"
+   last_bash_command="$BASH_COMMAND"
+   echo "Unexpected non-zero exit code $last_exit_code in ${BASH_SOURCE[*]} at line ${BASH_LINENO[*]} detected!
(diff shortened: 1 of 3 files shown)


zeha:
btw this is one of the things we could probably stop worrying about if we decide to base everything on top of mmdebstrap, which already supports extra chrooted commands etc.

PR #298, opened 2025-01-07
Mark packages as automatically installed

zeha:
This way a later "apt autoremove" will also remove library/leaf packages installed by the initial bootstrap.

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -83,6 +83,21 @@ askpass() {
 }
 # }}}
 
+# mark existing packages as automatically installed {{{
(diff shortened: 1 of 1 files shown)


mika:
> seems like this only works starting in bookworm. could skip it for older releases, maybe?

that sounds like a worthwhile idea, ack!

zeha:
> seems like this only works starting in bookworm. could skip it for older releases, maybe?

implemented that

mika (approved):

PR #301, opened 2025-01-14
Update /etc/timezone only if it exists

zeha:
tzdata 2024b-6 stops creating /etc/timezone, so we should not create it either.

https://salsa.debian.org/glibc-team/tzdata/-/commit/c071fbc167ad306394b576879d515e6ecbe5a734


jkirk (approved):
See: https://github.com/grml/grml-live/pull/286#pullrequestreview-2551112019

PR #303, opened 2025-01-20
Follow rename of grml-debian-keyring to grml-keyring

zeha:
https://github.com/grml/grml-debian-keyring/pull/9


mika (approved):
Please submit only once grml-keyring package is in at least grml-testing :)

PR #309, opened 2025-01-29
Use mmdebstrap as our only debootstrap implementation

zeha:
Simplify what we support.

TODO:
- [ ] remove DEBOOTSTRAP_OPTS, to totally encapsulate mmdebstrap
- [ ] fix --iso with mmdebstrap
- [ ] support --include and --exclude
- [ ] other commonly relevant DEBOOTSTRAP_OPTS?
- [ ] merge ISO and MIRROR code paths where possible?


adrelanos:
Could you please keep variable `"$DEBOOTSTRAP"`? Our distribution as a user of grml-debootstrap is using it.

```
   [ -n "$DEBOOTSTRAP" ] || export DEBOOTSTRAP="$mmdebstrap_wrapper"
```

This is [`mmdebstrap_wrapper`](https://github.com/Kicksecure/derivative-maker/blob/master/help-steps/pbuilder-debootstrap-command-filter). It's a wrapper around `mmdebstrap`.

mika:
> Could you please keep variable `"$DEBOOTSTRAP"`? Our distribution as a user of grml-debootstrap is using it.
> 
> ```
>    [ -n "$DEBOOTSTRAP" ] || export DEBOOTSTRAP="$mmdebstrap_wrapper"
> ```
> 
> This is [`mmdebstrap_wrapper`](https://github.com/Kicksecure/derivative-maker/blob/master/help-steps/pbuilder-debootstrap-command-filter). It's a wrapper around `mmdebstrap`.

We're trying to make grml-debootstrap more maintainable and more focused on being a tool rather than a construction kit.

Looking at your script, I don't really understand what's your use case exactly here, given the length of your wrapper, what purpose provides `grml-debootstrap` for you at all then, and what exactly is missing yet?

adrelanos:
We're using the same mmdebstrap wrapper for cowbuilder, pbuilder and grml-debootstrap.


```
mmdebstrap \
   --verbose \
   --debug \
   --format directory \
   --variant=required \
   --architectures="$dist_build_multiarch_package_item" \
   --aptopt="$dist_aptgetopt_file" \
   --customize-hook='echo "mmdebstrap hooks START"' \
   --customize-hook='echo "----------"' \
   --customize-hook='ls -la "$1" || true' \
   --customize-hook='realpath "$1" || true' \
   --customize-hook='echo "----------"' \
   --customize-hook='cat "$1/etc/apt/sources.list.d/0000debian_stable_current_clearnet.list" || true' \
   --customize-hook='cat "$1/etc/apt/apt.conf.d/99mmdebstrap" || true' \
   --customize-hook='echo "----------"' \
   --customize-hook='rm --force --verbose "$1/etc/apt/sources.list.d/0000debian_stable_current_clearnet.list"' \
   --customize-hook='rm --force --verbose "$1/etc/apt/apt.conf.d/99mmdebstrap"' \
   --customize-hook='echo "----------"' \
   --customize-hook='echo "mmdebstrap hooks END"' \
   $include_opt \
   "$dist_build_apt_stable_release" \
   "$BUILDPLACE" \
   "$dist_build_sources_list_primary"
```

We're injecting our own APT options file, various hook commands, a custom `--include` option and we're using a `sources.list` file instead of `https://` style `$MIRROR`.

The new grml-debootstrap code is essentially:

```
    mmdebstrap $ARCHCMD $DEBOOTSTRAP_OPT "$RELEASE" "$MNTPOINT" "$MIRROR"
```

Could we use `$DEBOOTSTRAP_OPT` instead? No. Not yet. That would need to become `"${DEBOOTSTRAP_OPT[@]}"`.

Aka "proper bash array syntax". Example:

```
cmd+=("cp")
cmd+=("-r")

"${cmd[@]}"
```

Related: `$DEBOOTSTRAP_OPT` is also currently under threat of removal in:

* https://github.com/grml/grml-debootstrap/issues/311

zeha:
Should add `--skip=check/empty` and disable cleanup of apt files

Related: https://github.com/grml/grml-debootstrap/issues/314#issuecomment-2624912601

mika (commented):
I like it! (Is set to draft, so don't know whether it's actually ready, but if you're fine with it, assume +2 from my side :)) Thx!

zeha:
TODOs are really in #311.

zeha:
> Could we use `$DEBOOTSTRAP_OPT` instead? No. Not yet. That would need to become `"${DEBOOTSTRAP_OPT[@]}"`.

Why?

It currently does word splitting.

zeha:
> Should add `--skip=check/empty` and disable cleanup of apt files
> 
> Related: [#314 (comment)](https://github.com/grml/grml-debootstrap/issues/314#issuecomment-2624912601)

Done!

adrelanos:
> > Could we use `$DEBOOTSTRAP_OPT` instead? No. Not yet. That would need to become `"${DEBOOTSTRAP_OPT[@]}"`.
> 
> Why? It currently does word splitting when calling mmdebstrap.

When using `'`, `"` and spaces, a proper array is likely required. (Maybe avoidable if using a lot of escaping.) But...

> (`DEBOOTSTRAP_OPT` will go, but for a little while it should just work as is.)

... not worth creating a test case to demonstrate since `DEBOOTSTRAP_OPT` is also unfortunately planned to be removed.

Since we'll unfortunately will require a fork anyhow (since both `$DEBOOTSTRAP` and `DEBOOTSTRAP_OPT`) are planned to be removed.

zeha:
> Since we'll unfortunately will require a fork anyhow (since both `$DEBOOTSTRAP` and `DEBOOTSTRAP_OPT`) are planned to be removed.

While reshaping grml-debootstrap is in progress, injecting your custom mmdebstrap via PATH should be possible.

adrelanos:
> , injecting your custom mmdebstrap via PATH should be possible.

Yes, that works.

Issue #311, opened 2025-01-29
Clarify design goals


zeha:
Need to clarify what grml-debootstrap should be, and which usecases it should support.

zeha:
Features we think which can or should go away:

- `--iso` (can) - uninteresting to us to take debian packages from an iso
- `DEBOOTSTRAP=...` (should)
- `DEBOOTSTRAP_OPTS=...` (should)
- ?

adrelanos:
As mentioned in https://github.com/grml/grml-debootstrap/pull/309#issuecomment-2624491579, for our Debian derivative distributions we need variable `DEBOOTSTRAP` and/or `DEBOOTSTRAP_OPTS` (as proper bash array).

zeha:
We also know that we need support for per-release package lists, as the available packages and useful defaults change per Debian release.

PR #312, opened 2025-01-30
Drop --iso support

zeha:
Installing using a loopback-mounted ISO as an unsigned mirror seems not very useful today.

Reference: https://github.com/grml/grml-debootstrap/issues/311

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -93,8 +93,6 @@ Usage: $PN [options]
 Bootstrap options:
 
   -m, --mirror <URL>     Mirror which should be used for apt-get/aptitude.
-  -i, --iso <mnt>        Mountpoint where a Debian ISO is mounted to, for use
-                         instead of fetching packages from a mirror.
   -r, --release <name>   Release of new Debian system (default: bullseye).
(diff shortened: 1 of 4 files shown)


grml-debootstrap
```diff
@@ -879,10 +863,10 @@ prompt_for_mirror()
...
 
   else # CHOOSE_MIRROR == local
-     [ -n "$ISO" ] || ISO='/mnt/mirror'
-     ISO="$(dialog --stdout --title "${PN}" --inputbox \
-               "Please enter directory name you would like to use for installing packages." \
-               0 0 "$ISO")" || bailout
+     MIRROR='file:///mnt/mirror'
+     MIRROR="$(dialog --stdout --title "${PN}" --inputbox \
+               "Please enter the directory in file:/// URL form you would like to use for installing packages." \
+               0 0 "$MIRROR")" || bailout
```
mika, line 869: this is the only unclear issue for me that we might wanna verify, other than this LGTM (probably not worth a debian/NEWS entry?)

zeha: removed this part

mika (approved):

Issue #310, opened 2025-01-29
use `LC_ALL=C` instead of `LANG=C`


adrelanos:
Based on https://unix.stackexchange.com/questions/576701/what-is-the-difference-between-lang-c-and-lc-all-c. Specifically reply https://unix.stackexchange.com/a/576711/49297. Quote:

> `LANG` does not override:
> 
> ```
> $ export LC_NUMERIC="de_DE.UTF-8"
> $ export LANG=C
> $ printf "%.2f\n" 3.14
> -bash: printf: 3.14: invalid number
> 3,00
> ```
> 
> But `LC_ALL` does:
> 
> ```
> $ export LC_NUMERIC="de_DE.UTF-8"
> $ export LC_ALL=C
> $ printf "%.2f\n" 3.14
> 3.14
> ```

Confirmed.

Therefore mentions of `LANG=C` should be replaced with `LC_ALL=C`, where applicable.

If you agree with this, I am happy to send a pull request.

mika:
Ah interesting catch, thanks, yes that sounds reasonable!

PR #313, opened 2025-01-30
use LC_ALL=C instead of LANG=C

adrelanos:
fixes https://github.com/grml/grml-debootstrap/issues/310

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -715,7 +715,7 @@ Do you want to continue installing Debian using this frontend?" 0 0 || bailout 0
 # ask for target {{{
 prompt_for_target()
 {
-  AVAILABLE_PARTITIONS=$(LANG=C fdisk -l 2>/dev/null | \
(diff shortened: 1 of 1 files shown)


mika (commented):
LGTM, just checking:

```
clean_chroot() {
  # inside the chroot system locales might not be available, so use minimum:
  local -a env_vars=(
    "LANG=C"
    "LC_ALL=C"
    "LANGUAGE=C"
[....]
```

This can stay as is?

adrelanos:
That's legacy carried forward from ages ago. In theory, it can be reduced to `LC_ALL=C`, but I didn't test that yet and wanted to split this into a simpler change and left this as future work.

mika:
> That's legacy carried forward from ages ago. In theory, it can be reduced to `LC_ALL=C`, but I didn't test that yet and wanted to split this into a simpler change and left this as future work.

Ok thx!

PR #319, opened 2025-01-31
use `LC_ALL=C` instead of `LANG=C` inside chroot

adrelanos:
no longer set redundant environment variables LANG=C and LANGUAGE=C in because already setting LC_ALL=C

https://github.com/grml/grml-debootstrap/issues/310
https://github.com/grml/grml-debootstrap/pull/313


mika (changes requested):
s/no longer set redudant envrionment variables/no longer set redundant environment variables/ please

adrelanos:
Done.

Issue #320, opened 2025-01-31
Remove Grub code duplication


zeha:
We have grub-install code in `chroot-script` and in `grml-debootstrap`. Also we have lots of duplication between ARM, VM, and bare-metal x86.  None of that seems wise.

adrelanos:
> Also we have lots of duplication between ARM, VM, and bare-metal x86.

* https://github.com/grml/grml-debootstrap/pull/299 is refactoring that.

Related:

* https://github.com/grml/grml-debootstrap/issues/258

PR #331, opened 2025-03-21
Deduplicate GRUB installation code

ArrayBolt3:
Previously, we had two separate implementations of GRUB installation code, one that worked for physical disks and one that worked for virtual disks. The physical disk code works for virtual disks with some minor changes, so make those changes and discard the virtual disk bootloader installer.

Fixes https://github.com/grml/grml-debootstrap/issues/320

Needs intensive testing before merging.

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -648,6 +648,8 @@ device_to_id() {
     echo "$1"
     return 0
   fi
+
+  echo "ERROR: Device '$1' does not exist!"
   return 1
 }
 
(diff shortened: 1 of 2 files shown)


ArrayBolt3:
I don't think the ARM builds are failing because of the PR, looks like a CI bug...

zeha:
> I don't think the ARM builds are failing because of the PR, looks like a CI bug...

To me it looks like the PR does break arm.

ArrayBolt3:
argh, you're right, I misread the logs and thought it was failing to sign into a builder machine, didn't realize that was the test step. Will reproduce locally and find out why.

ArrayBolt3:
Should be fixed now. I think the issue was that the GRUB bootloader wasn't being installed to the fallback location anymore. I was able to do a successful, bootable-out-of-the-box ARM64 build with the latest commit.

zeha (approved):

mika (approved):
Untested from my side, but LGTM! thx

PR #333, opened 2025-03-31
efi_support: drop BIOS-mode info

zeha:
Absence of an EFI runtime does not prevent loading efivarfs since commit
torvalds/linux@301de9a2055357375a4e1053d9df0f8f3467ff00 which landed in Linux v6.3.

Thus our check code cannot work anymore, and the message is at best misleading.

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -977,11 +977,6 @@ format_efi_partition() {
 # check for EFI support or try to enable it {{{
 efi_support() {
   local efivars_loaded=false
-  # this is for kernels versions before v3.10, which didn't provide efivarfs yet
-  if modprobe efivars &>/dev/null ; then
-    efivars_loaded=true
-  fi
-  # kernel versions v3.10 and newer usually provide efivarfs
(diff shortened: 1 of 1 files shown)


grml-debootstrap
```diff
@@ -976,26 +976,17 @@ format_efi_partition() {
...
-    efivars_loaded=true
-  fi
-  # kernel versions v3.10 and newer usually provide efivarfs
-  if modprobe efivarfs &>/dev/null ; then
-    efivars_loaded=true
-  fi
+  # Absence of an EFI runtime does not prevent loading efivarfs since commit
+  # 301de9a2055357375a4e1053d9df0f8f3467ff00 which landed in Linux v6.3.
+  # Unconditionally load it, in case nothing else attempted it.
+  modprobe efivarfs &>/dev/null
```
mika, line 982: This still *might* fail, nor? If so, then maybe add `|| true` as elsewhere, since we're running under `set -eE`?

zeha: Indeed. Fixed

mika (approved):

Issue #219, opened 2023-10-19
avoid unicode in source code if reasonably avoidable


adrelanos:
Why? Security.

References:
https://www.kicksecure.com/wiki/Unicode (written by me)
https://trojansource.codes/

For `config` that is trivial and probably even non-controversial.

Created https://github.com/grml/grml-debootstrap/pull/218 for it. In the `config` the unicode seemed to be a bug and not useful at all.

The only other file with unicode is `debian/changelog`.

    grep -P -n '[^[:ascii:]]' debian/changelog

```
21:  The "¡hola bookworm!" release
80:  * [457e3e4] config: Replace em dash — by `--` for switch in comment
148:  [ Antoine Beaupré ]
444:  The "monster kill ▬█████████" release
527:  The "jessie partyyyyyy ♫♫♫" release
563:  The "[✔] ready for jessie" release
631:  The "happy easter ＼(◎o◎)／" release
1365:    Thanks for the patch, Alexander 'Stone' Steinböck!
```

But these are not as non-controversial to remove. Butchering real names is probably unwanted. And maybe less important because it's a textual file. Not code that is being interpreted.

mika:
ACK, thanks for the references and also the PR, merged already! :)

PR #334, opened 2025-04-02
Replace inline ESC with printf calls

adrelanos:
https://github.com/grml/grml-debootstrap/issues/219

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -180,10 +180,10 @@ if [ "${TERM}" = "dumb" ] ; then
   WARN=
   NORMAL=
 else
-  GOOD='[32;01m'
-  BAD='[31;01m'
-  WARN='[33;01m'
-  NORMAL='[0m'
+  GOOD="$(printf '\033[32;01m')"
+  BAD="$(printf '\033[31;01m')"
+  WARN="$(printf '\033[33;01m')"
+  NORMAL="$(printf '\033[0m')"
 fi
 
 einfo() {


adrelanos:
I wrote a test script to make sure it is really the same. Just a different notation.

```
#!/bin/bash

set -e

GOOD1="$(printf '\033[32;01m')"
BAD1="$(printf '\033[31;01m')"
WARN1="$(printf '\033[33;01m')"
NORMAL1="$(printf '\033[0m')"

GOOD='[32;01m'
BAD='[31;01m'
WARN='[33;01m'
NORMAL='[0m'

if [ ! "$GOOD" = "$GOOD1" ]; then
  exit 1
fi

if [ ! "$BAD" = "$BAD1" ]; then
  exit 1
fi

if [ ! "$WARN" = "$WARN1" ]; then
  exit 1
fi

if [ ! "$NORMAL" = "$NORMAL1" ]; then
  exit 1
fi

echo "$GOOD1 test $NORMAL"
echo "$BAD test $NORMAL"
echo "$WARN test $NORMAL"
echo "$NORMAL test $NORMAL"

exit 0
```

mika (commented):
Thx! LGTM, would appreciate second pair of eyes though :)

zeha:
Please retitle to:
`Replace inline ESC with printf calls`

"unicode" is misleading IMO.

adrelanos:
New title applied.

zeha:
Did a squash merge to avoid another back-and-forth, but I really did mean the commit message, not the PR title.

adrelanos:
Appreciated, thank you!

PR #335, opened 2025-04-05
Add Raspberry Pi image generation support

### LINKED ISSUE #114: Rasberry PI (RPI) support
It's possible.
@Algernon-01 managed to implement it in @Whonix which is using grml-debootstrap. (https://github.com/Whonix/Whonix/pull/419)

In essence if I did not overlook something - deduction from theory (code reading):

* `export DEBOOTSTRAP='qemu-debootstrap'`
* `export FSCK='no'` (?) (Maybe @Algernon-01 can explain?)
* skip installation of grub (since it does not exist for that platform) - https://github.com/grml/grml-debootstrap/pull/113
* packages ((https://github.com/Whonix/anon-meta-packages/pull/11)): [rpi-patches](https://github.com/Algernon-01/rpi-patches), wpasupplicant, firmware-iwlwifi, firmware-atheros, firmware-brcm80211, firmware-ralink, firmware-realtek, iw, fake-hwclock
* parted, mkfs stuff as per [2375_build_rpi_fs](https://github.com/Algernon-01/Whonix/blob/master/build-steps.d/2375_build_rpi_fs)

Would be cool if grml-debootstrap supported RPI out of the box.

ArrayBolt3:
Adds Raspberry Pi installation and image generation support. Cross-builds are supported (you can create a working Raspberry Pi image on an amd64 machine).

Some notes:

* raspi-firmware generates the `cmdline.txt` for the kernel in a highly non-ideal way, using a device path rather than a UUID for identifying the root directory. This has to be manually patched to get the image to boot at all.
* If your host system is a Pi using one of the raspi.debian.net images, you probably want to pass `--defaultinterfaces` to grml-debootstrap, since the `/etc/network/interfaces` file used in the raspi.debian.net images is just a pointer to `/etc/network/interfaces.d`, which grml-debootstrap doesn't copy over. Using `--defaultinterfaces` will prevent this from being a problem, not using it will mean you'll have no network when you boot the system.
* The image **DOES NOT** automatically resize itself on first boot. The RPi image generation is more like VM image generation, the image is fixed-size and you have to resize it afterwards. If you don't want that, you can install directly to a USB drive or SD card.
* This uses the current way of doing things as implemented in the normal Debian Raspberry Pi images, where the kernel is booted directly by the firmware bootloader. There are lots of other ways to boot a Pi (involving u-boot and/or edkII), which might be more suitable for some use cases. However, as grml-debootstrap is intended to install pure Debian, I tried to stick with a pure Debian approach as much as possible. I've also done research showing that a direct-kernel-boot install like this can be converted to a U-Boot + GRUB install without too much effort. (See https://www.kicksecure.com/wiki/Dev/boot#Booting_Debian_Trixie_with_GRUB_.2B_u-boot_on_Raspberry_Pi_4)

Fixes https://github.com/grml/grml-debootstrap/issues/114

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -75,7 +75,6 @@ MNTPOINT="/mnt/debootstrap.$$"
 [ -n "$TIMEZONE" ] || TIMEZONE='Europe/Vienna'
 [ -n "$TUNE2FS" ] || TUNE2FS='tune2fs -c0 -i0'
 [ -n "$UPGRADE_SYSTEM" ] || UPGRADE_SYSTEM='yes'
-[ -n "$VMSIZE" ] || VMSIZE="2G"
 [ -n "$GRUB_INSTALL" ] || GRUB_INSTALL='yes'
 
 # make sure interactive mode is only executed when
@@ -102,10 +101,19 @@ Bootstrap options:
       --nodebootstrap    Skip debootstrap, only do configuration to the target.
       --grub <device>    Target for grub installation. Usage example: /dev/sda
       --efi <device>     Target for EFI installation. Usage example: /dev/sda1
+      --rpifirm <device> Target for Raspberry Pi firmware installation. Usage
+                         example: /dev/mmcblk0p1
       --arch <arch>      Set target architecture.
       --filesystem <fs>  Filesystem that should be used when target is a partition
                          or Virtual Machine (see --vmfile).
       --force            Do not prompt for user acknowledgement.
+
+Options for image file deployments:
+
+      --imagesize        Use specified size for size of image file (default:
+                         2G for BIOS and UEFI systems, 3G for Raspberry Pi
+                         images). Syntax as supported by qemu-img, like:
+                         --imagesize 3G
 
 Options for Virtual Machine deployment:
 
@@ -116,9 +124,17 @@ Options for Virtual Machine deployment:
       --vmfile           Like --vm, but install into a regular file (created by
                          'qemu-img create -f raw ...') instead.
                          Example: --vmfile --target /mnt/sda1/qemu.img
-      --vmsize <size>    Use specified size for size of VM file (default: 2G).
-                         Syntax as supported by qemu-img, like: --vmsize 3G
+      --vmsize <size>    Deprecated alias for --imagesize.
       --vmefi            Create an EFI boot partition for the VM.
+
+Options for Raspberry Pi deployment:
+
+      --rpi              Set up a Raspberry Pi image on an existing block device
+                         instead of plainly installing to a partition or directory.
+                         Needs to be combined with --target.
+                         Example: --rpi --target /dev/mapper/your-rpi-disk
+      --rpifile          Like --rpi, but install into a regular file (created by
+                         'qemu-img create -f raw ...') instead.
 
 Configuration options:
 
@@ -369,7 +385,7 @@ fi
 # }}}
 
 # cmdline handling {{{
-CMDLINE_OPTS=mirror:,iso:,release:,target:,mntpoint:,debopt:,defaultinterfaces,interactive,nodebootstrap,nointerfaces,nokernel,nopackages,filesystem:,config:,confdir:,packages:,chroot-scripts:,scripts:,post-scripts:,pre-scripts:,debconf:,vm,vmfile,vmsize:,vmefi,keep_src_list,hostname:,password:,nopassword,grmlrepos,backportrepos,bootappend:,grub:,efi:,arch:,insecure,verbose,help,version,force,debug,contrib,non-free,remove-configs,sshcopyid,sshcopyauth
+CMDLINE_OPTS=mirror:,iso:,release:,target:,mntpoint:,debopt:,defaultinterfaces,interactive,nodebootstrap,nointerfaces,nokernel,nopackages,filesystem:,config:,confdir:,packages:,chroot-scripts:,scripts:,post-scripts:,pre-scripts:,debconf:,vm,vmfile,vmsize:,imagesize:,vmefi,rpi,rpifile,keep_src_list,hostname:,password:,nopassword,grmlrepos,backportrepos,bootappend:,grub:,efi:,rpifirm:,arch:,insecure,verbose,help,version,force,debug,contrib,non-free,remove-configs,sshcopyid,sshcopyauth
 
 if ! _opt_temp=$(getopt --name grml-debootstrap -o +m:i:r:t:p:c:d:vhV --long \
   $CMDLINE_OPTS -- "$@"); then
@@ -391,6 +407,9 @@ while :; do
   --target|-t)         # Target partition (/dev/...) or directory
     shift; _opt_target="$1"
     ;;
+  --imagesize)         # size of image deployment file
+    shift; _opt_imagesize="$1"
+    ;;
   --vm)                # Virtual machine image (no file)
     _opt_vm="T"
     ;;
@@ -398,10 +417,16 @@ while :; do
     _opt_vmfile="T"
     ;;
   --vmsize)            # size of Virtual machine file
-    shift; _opt_vmsize="$1"
+    shift; _opt_imagesize="$1"
     ;;
   --vmefi)            # Create an EFI boot partition for the VM
     _opt_vmefi="T"
+    ;;
+  --rpi)               # RPi image (no file)
+    _opt_rpi="T"
+    ;;
+  --rpifile)           # RPi image file
+    _opt_rpifile="T"
     ;;
   --mntpoint|-p)       # Mountpoint used for mounting the target system
     shift; _opt_mntpoint="$1"
@@ -503,6 +528,9 @@ while :; do
   --efi)               # Target for EFI boot installation
     shift; _opt_efi="$1"
     ;;
+  --rpifirm)           # Target for RPi firmware installation
+    shift; _opt_rpifirm="$1"
+    ;;
   --contrib)           # Add 'contrib' to list of components
     _opt_contrib=T
     ;;
@@ -565,9 +593,11 @@ done
 [ "$_opt_release" ]             && RELEASE=$_opt_release
 [ "$_opt_target" ]              && TARGET=$_opt_target
 [ "$_opt_vm" ]                  && VIRTUAL=1
-[ "$_opt_vmfile" ]              && VMFILE=1 && VIRTUAL=1
-[ "$_opt_vmsize" ]              && VMSIZE=$_opt_vmsize
+[ "$_opt_vmfile" ]              && IMAGEFILE=1 && VIRTUAL=1
+[ "$_opt_imagesize" ]           && IMAGESIZE=$_opt_imagesize
 [ "$_opt_vmefi" ]               && VMEFI=1
+[ "$_opt_rpi" ]                 && RPI=1
+[ "$_opt_rpifile" ]             && IMAGEFILE=1 && RPI=1
 [ "$_opt_mntpoint" ]            && MNTPOINT=$_opt_mntpoint
 [ "$_opt_debopt" ]              && DEBOOTSTRAP_OPT=$_opt_debopt
 [ "$_opt_interactive" ]         && INTERACTIVE=1
@@ -593,6 +623,7 @@ done
 [ "$_opt_bootappend" ]          && BOOT_APPEND=$_opt_bootappend
 [ "$_opt_grub" ]                && GRUB=$_opt_grub
 [ "$_opt_efi" ]                 && EFI=$_opt_efi
+[ "$_opt_rpifirm" ]             && RPIFIRM=$_opt_rpifirm && RPI=1
 [ "$_opt_arch" ]                && ARCH=$_opt_arch
 [ "$_opt_insecure" ]            && echo "Warning: --insecure is deprecated, continuing anyway."
 [ "$_opt_force" ]               && FORCE=$_opt_force
@@ -604,6 +635,35 @@ done
 [ -z "$COMPONENTS" ]            && COMPONENTS="main"
 [ "$_opt_contrib" ]             && COMPONENTS="$COMPONENTS contrib"
 [ "$_opt_non_free" ]            && COMPONENTS="$COMPONENTS non-free-firmware non-free"
+
+# If $VMSIZE is set and $IMAGESIZE is unset, put $VMSIZE's value in $IMAGESIZE.
+#
+# If $VMSIZE is unset and $IMAGESIZE is set, keep $IMAGESIZE's value.
+#
+# If $VMSIZE is unset and $IMAGESIZE is unset, $IMAGESIZE defaults to 2G for VM images and 3G for
+# RPi images (which need a bit more space).
+if [ -n "$VMSIZE" ] && [ -z "$IMAGESIZE" ]; then
+  IMAGESIZE="$VMSIZE"
+elif [ -z "$IMAGESIZE" ]; then
+  if [ "$RPI" ]; then
+    IMAGESIZE="3G"
+  else
+    IMAGESIZE="2G"
+  fi
+fi
+
+if [ "$RPI" ] && [ -z "$ARCH" ]; then
+  # Ensure the architecture for the build is set to arm64 even if building on
+  # a non-arm64 machine, this is necessary for cross-builds to work
+  ARCH='arm64'
+fi
+
+# RPi images *require* non-free-firmware as the raspi-firmware package resides
+# there, and Raspberry Pis are not bootable without it.
+if [ -z "$_opt_non_free" ] && [ "$RPI" ]; then
+  eerror "Raspberry Pi images require non-free firmware, please add --non-free to your command line."
+  bailout 1
+fi
 
 # command line option checks
 if [ "$_opt_scripts_set" ] ; then
@@ -630,6 +690,27 @@ fi
 if [ "$_opt_efi" ] && [ "$_opt_vmefi" ] ; then
   eerror "The --efi option is incompatible with --vmefi, please drop one of them from your command line."
   eerror "--vmefi should be used if the VM image should be self-bootable. Use --efi if you need a secondary disk to act as a boot disk."
+  bailout 1
+fi
+
+if [ "$RPI" ] && [ "$EFI" ] ; then
+  eerror "Raspberry Pi UEFI support has not yet been implemented, please drop EFI-related options from your command line."
+  bailout 1
+fi
+
+if [ "$RPI" ] && [ "$VIRTUAL" ] ; then
+  eerror "Raspberry Pi and virtual machine image options are mutually exclusive, please drop one of them from your command line."
+  eerror "If you want to build a virtual machine image for use on a Raspberry Pi, please build a VM image for the ARM64 architecture."
+  bailout 1
+fi
+
+if [ "$RPI" ] && [ "$GRUB" ] ; then
+  eerror "Raspberry Pi images do not use GRUB, please drop GRUB-related options from your command line."
+  bailout 1
+fi
+
+if [ "$RPI" ] && [ -n "$ARCH" ] && [ "$ARCH" != 'arm64' ] ; then
+  eerror "Raspberry Pi images support only the arm64 architecture. Please drop the --arch option from your command line or change it to specify arm64."
   bailout 1
 fi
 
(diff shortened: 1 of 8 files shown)


ArrayBolt3:
Found a bug in grml-debootstrap that existed before this PR - installing a VM image to a device such as `/dev/sda` was impossible because grml-debootstrap was trying to format devices named `/dev/mapper/sda1p1` (which is wrong for multiple reasons). This was because it always assumed that device names were formatted as `{disk_driver}{disk_number}p{partition_number}`, when some devices are formatted as `{disk_driver}{disk_letter}{partition_number}`. Fixed in an added commit (https://github.com/grml/grml-debootstrap/pull/335/commits/752a6114ee5f82a34177b2bde7622115b70213ac), that one should be able to be cherry-picked without merging this entirely.

grml-debootstrap
```diff
@@ -75,7 +75,7 @@ MNTPOINT="/mnt/debootstrap.$$"
 [ -n "$TIMEZONE" ] || TIMEZONE='Europe/Vienna'
 [ -n "$TUNE2FS" ] || TUNE2FS='tune2fs -c0 -i0'
 [ -n "$UPGRADE_SYSTEM" ] || UPGRADE_SYSTEM='yes'
-[ -n "$VMSIZE" ] || VMSIZE="2G"
+#[ -n "$IMAGESIZE" ] || IMAGESIZE="2G"
```
zeha, line 78: ?

grml-debootstrap
```diff
@@ -613,6 +645,27 @@ case "${RELEASE}" in
...
+  if [ "$RPI" ]; then
+    IMAGESIZE="3G"
+  else
+    IMAGESIZE="2G"
+  fi
+fi
+
+# RPi images *require* non-free-firmware as the raspi-firmware package resides
+# there, and Raspberry Pis are not bootable without it.
+if [ -z "$_opt_non_free" ] && [ "$RPI" ]; then
```
zeha, line 658: i think this will need a warning, or better it should abort if non-free(-firmware) was not requested

grml-debootstrap
```diff
@@ -951,26 +1025,29 @@ fi
...
-  fi
-
-  if fsck.vfat -bn "$EFI" >/dev/null; then
-    einfo "EFI partition $EFI seems to have a FAT filesystem, not modifying."
-  else
-    einfo "EFI partition $EFI doesn't seem to be formatted, creating filesystem."
-    if ! mkfs.fat -F32 -n "EFI" "$EFI" ; then
-      eerror "Error while creating filesystem on ${EFI}."
+# format boot info partitions (ESP, Raspberry Pi firmware) {{{
+format_bootinfo_partitions() {
```
zeha, line 1029: other tools tend to call them "firmware boot partitions"

grml-debootstrap
```diff
@@ -1469,6 +1555,9 @@ prepare_vm() {
     else
       LOOP_PART="${LOOP_PART%p1}p3"
     fi
+  elif [ -n "$RPI" ]; then
+    export RPIFIRM_TARGET="/dev/mapper/$LOOP_PART" # '/dev/mapper/loop0p1'
```
zeha, line 1565: i have my doubts if this needs `export`

grml-debootstrap
```diff
@@ -1732,6 +1825,11 @@ iface ${interface} inet dhcp
     einfo "Copying /etc/network/interfaces from host to target system"
     mkdir -p "${MNTPOINT}/etc/network"
     cp ${VERBOSE:+-v} /etc/network/interfaces "${MNTPOINT}/etc/network/interfaces"
+    if [ -r /etc/network/interfaces.d ] ; then
```
zeha, line 1838: this should land in a separate PR

zeha (commented):
some initial comments, without any further consideration

zeha:
> Found a bug in grml-debootstrap that existed before this PR - installing a VM image to a device such as `/dev/sda` was impossible because grml-debootstrap was trying to format devices named `/dev/mapper/sda1p1` (which is wrong for multiple reasons). This was because it always assumed that device names were formatted as `{disk_driver}{disk_number}p{partition_number}`, when some devices are formatted as `{disk_driver}{disk_letter}{partition_number}`. Fixed in an added commit ([752a611](https://github.com/grml/grml-debootstrap/commit/752a6114ee5f82a34177b2bde7622115b70213ac)), that one should be able to be cherry-picked without merging this entirely.

AFAIK, VM **images** were never supposed to be written to devices by grml-debootstrap. @mika ?

ArrayBolt3:
@zeha By installing a VM image directly to a device, I mean using a construct such as `--target /dev/sda --vm`, not using `--vmfile`. I probably just used the wrong terminology.

mika:
> @zeha By installing a VM image directly to a device, I mean using a command line such as `--target /dev/sda --vm`, not using `--vmfile`. I probably just used the wrong terminology.

Hm, what's the use case for `--target /dev/sda --vm` exactly, as in: why install an VM into a block device at all and not into a plain file only? (Not claiming that there's no use case for it, but to clarify actual needs and how far we should go with supporting that :))

ArrayBolt3:
Off the top of my head, I know people oftentimes store VMs in LVM volumes
rather than in files. I don't know for sure why, but I suspect performance
plays a role.

For the Raspberry Pi case specifically, being able to install directly to a
physical device would be useful to prevent someone from having to generate
an image, flash the image, and then resize the image to occupy the entire
disk. They can just install straight to the device and use it.

mika:
> Off the top of my head, I know people oftentimes store VMs in LVM volumes rather than in files. I don't know for sure why, but I suspect performance plays a role. For the Raspberry Pi case specifically, being able to install directly to a physical device would be useful to prevent someone from having to generate an image, flash the image, and then resize the image to occupy the entire disk. They can just install straight to the device and use it.

Ah I see, thx4info!

zeha:
JFTR, this is even documented in man grml-debootstrap:

>       --vm
>           Set up a Virtual Machine on an existing block device, which will be partitioned. This
>           allows deployment of a Virtual Machine. The options needs to be combined with the
>           --target option. This option automatically enables the --defaultinterfaces option. 
>           Usage example: --vm --target /dev/mapper/your-vm-disk

(Note the example.)

grml-debootstrap
ArrayBolt3: Was shuffling code around and left "old code" commented out to avoid making destructive changes mid-work. I should remove that line before taking this out of draft status.

grml-debootstrap
ArrayBolt3: Currently there isn't an option for just enabling non-free-firmware but not the entirety of non-free. Should that be implemented, or is the job `--non-free` already does good enough?

grml-debootstrap
ArrayBolt3: It's used in a similar line for the EFI system partition a little bit above: `export EFI_TARGET="/dev/mapper/$LOOP_PART" # '/dev/mapper/loop0p1'` Maybe the export can be removed from both places?

grml-debootstrap
ArrayBolt3: Good point, this is sort of a quirk of the fact that I'm using one of the raspi.debian.net images as my host system and it stores its configuration in interfaces.d, not directly related to the kind of image I'm building.

grml-debootstrap
ArrayBolt3: Renamed.

grml-debootstrap
ArrayBolt3: Removed.

grml-debootstrap
ArrayBolt3: Added an error message with bailout if `--non-free` isn't specified. If we want something that only enables non-free-firmware when possible, I'd be happy to add that too.

grml-debootstrap
ArrayBolt3: Removed export both from this and from the `EFI_TARGET` bit, seems that things still work right in my testing.

grml-debootstrap
ArrayBolt3: Removed for now, will make a second PR for it.

ArrayBolt3:
Ran thorough tests to ensure things worked as expected. "Working" is defined as "the build doesn't error out, the produced image is bootable without having to take any special steps, the `root` user account can be logged into, the contents of /boot are appropriate for the installation being tested, the partition layout is appropriate for the installation being tested, and `apt update` successfully downloads an updated package database, ensuring network connectivity works. It may be required to modify /etc/network/interfaces to get the interface name right, this is expected."

* Builder hardware: x86_64 VM
  * `[x]` Can create a working x86_64 BIOS VM image? `sudo grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --arch amd64 --vmfile --target ./x86_64-bios-bookworm.img`
  * `[x]` Can create a working x86_64 UEFI VM image? `sudo grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --arch amd64 --vmfile --vmefi --target ./x86_64-efi-bookworm.img`
  * `[x]` Can create a working x86_64 BIOS VM direct install? `sudo grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --arch amd64 --vm --target /dev/vdb`
  * `[x]` Can create a working x86_64 UEFI VM direct install? `sudo grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --arch amd64 --vm --vmefi --target /dev/vdb`
  * `[x]` Can create a working arm64 UEFI VM image? `sudo grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --arch arm64 --vmfile --vmefi --target ./arm64-efi-bookworm.img`
  * `[x]` Can create a working arm64 UEFI VM direct install? `sudo grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --arch arm64 --vm --vmefi --target /dev/vdb`
  * `[x]` Can create a working RPi image? `sudo grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --rpifile --non-free --target ./arm64-rpi-bookworm.img`
  * `[x]` Can create a working RPi direct install? `sudo grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --rpi --non-free --target /dev/vdb`
* Builder hardware: arm64 Raspberry Pi 4
  * `[x]` Can create a working arm64 UEFI VM image? `grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --arch arm64 --vmfile --vmefi --target ./arm64-efi-bookworm.img`
  * `[x]` Can create a working arm64 UEFI VM direct install? `grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --arch arm64 --vm --vmefi --target /dev/sda`
  * `[x]` Can create a working RPi image? `grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --rpifile --non-free --target ./arm64-rpi-bookworm.img`
  * `[x]` Can create a working RPi direct install? `grml-debootstrap --release bookworm --password 'x' --force --defaultinterfaces --rpi --non-free --target /dev/sda`

I did fix a couple of bugs in the original PR as I went, but both fixes required small, non-invasive changes.

ArrayBolt3:
Kicksecure isn't really thrilled with the boot process that raspi-firmware uses by default, due to its effects on things like booting into older kernels, and customizing kernel parameters at boot time. If possible, we'd like to be using U-Boot and grub-efi-arm64 instead. I filed some feature requests upstream (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1102607, https://salsa.debian.org/raspi-team/image-specs/-/issues/78) to see if this can be changed. If things do change in that area and GRUB + U-Boot booting for the Pi is officially supported in Debian, could grml-debootstrap use that boot flow instead, or at least offer it as an option? (Obviously we're more than happy to send the patch for this, I just want to make sure this will be acceptable in the future so we can do project planning.)

ArrayBolt3:
@mika, @zeha Gentle ping, anything else that needs done to this?

mika:
> @mika, @zeha Gentle ping, anything else that needs done to this?

I'm personally not really a huge fan of Raspberry devices, but I assume there are users out there who might benefit from it. So IMO it's also about how intrusive this change would become, and how we can ensure that things keep working.
Looking at e.g. https://github.com/ptrsr/pi-ci, is this something we could somehow also provide through our GitHub CI? :thinking:

ArrayBolt3:
I'm not sure, I'm not exactly handy with Docker, and based on the documentation it looks like pi-ci is both an emulator and image generator, there doesn't appear to be a documented easy way to boot a custom image with it. Newer versions of QEMU have support for emulating the Raspberry Pi 4B, which might be able to be used somehow, but even then you have to specify the kernel and initramfs to boot directly (QEMU doesn't emulate the boot firmware's behavior of reading config.txt and figuring out what to boot from that). I guess it would be possible to kinda sorta emulate the boot firmware by mounting the build image, finding config.txt, parsing out the kernel and initramfs lines, copying those out of the image, then unmounting the image and booting the kernel + initramfs + image with QEMU. I'd be willing to give that a shot, though be warned that I'm not particularly experienced with CI systemd either :P

mika:
> That sounds rather complex and may not be very useful as it would not test on the CI if the image is actually bootable. It would test kernel + initramfs and anything after that. Not sure how useful that would be?
> 
> Is CI testing mandatory?

What we like and appreciate about CI testing is, that we catch any possible side-effects/breakages with any changes within g-d very early. It definitely improves our options to redesign/rewrite things, as well as accept contributions.
It's not a 100% "must have" for everything all the time, but it's something we definitely prefer having. :)

zeha:
> > That sounds rather complex and may not be very useful as it would not test on the CI if the image is actually bootable. It would test kernel + initramfs and anything after that. Not sure how useful that would be?
> > Is CI testing mandatory?
> 
> What we like and appreciate about CI testing is, that we catch any possible side-effects/breakages with any changes within g-d very early. It definitely improves our options to redesign/rewrite things, as well as accept contributions. It's not a 100% "must have" for everything all the time, but it's something we definitely prefer having. :)

To add my thought: everything that's not CI tested _will_ break. When we'll do a refactor, nobody will test stuff that's not automatically tested, and it will just be broken. My personal view is that it's close to mandatory.

mika:
> > > That sounds rather complex and may not be very useful as it would not test on the CI if the image is actually bootable. It would test kernel + initramfs and anything after that. Not sure how useful that would be?
> > > Is CI testing mandatory?
> > 
> > 
> > What we like and appreciate about CI testing is, that we catch any possible side-effects/breakages with any changes within g-d very early. It definitely improves our options to redesign/rewrite things, as well as accept contributions. It's not a 100% "must have" for everything all the time, but it's something we definitely prefer having. :)
> 
> To add my thought: everything that's not CI tested _will_ break. When we'll do a refactor, nobody will test stuff that's not automatically tested, and it will just be broken. My personal view is that it's close to mandatory.

*nod* && ACK!

ArrayBolt3:
kk, then I guess I'll be adding CI to this.

ArrayBolt3:
Unfortunately, it looks like CI is going to end up being very slow if we want to use QEMU's RPi4 emulation. QEMU is unable to use KVM when emulating a Raspberry Pi 4B because the CPU in the RPi supports EL3, which can be emulated but not virtualized using KVM and QEMU. https://gitlab.com/qemu-project/qemu/-/issues/1073 has some more info on this. It is possible to boot a generated Raspberry Pi image using a more VM-friendly machine type, but it doesn't result in as accurate of Raspberry Pi emulation, which might be undesirable.

Should I use the faster but less accurate mode, or the slower, more accurate one?

ArrayBolt3:
I'll go ahead and implement this using the faster but less accurate virtualization. It's all that's necessary to make sure the image was built properly, the accurate mode is probably only necessary for making sure the kernel or firmware itself is functional on the official hardware itself (which isn't what we need to test, that should be being tested by other projects).

ArrayBolt3:
Raspberry PI image CI testing added. https://github.com/ArrayBolt3/grml-debootstrap/actions/runs/14696085489 will show the results once the tests are done running. (I got them to pass fully in a separate branch first.)

ArrayBolt3:
@mika, @zeha, gentle ping, this should be ready for review. RPi image CI tests have been added and are passing.

tests/test-vm.sh
```diff
@@ -72,17 +73,30 @@ declare -a qemu_command
...
     cp /usr/share/AAVMF/AAVMF_VARS.fd efi_vars.fd
     qemu_command=( qemu-system-aarch64 )
     qemu_command+=( -machine "type=virt,gic-version=max,accel=kvm:tcg" )
     qemu_command+=( -drive "if=pflash,format=raw,unit=0,file.filename=/usr/share/AAVMF/AAVMF_CODE.no-secboot.fd,file.locking=off,readonly=on" )
     qemu_command+=( -drive "if=pflash,format=raw,unit=1,file=efi_vars.fd" )
+  fi
 else
-    echo "E: unsupported ${DPKG_ARCHITECTURE}"
-    exit 1
+  echo "E: unsupported ${DPKG_ARCHITECTURE}"
```
mika, line 98: same (redirect error message to stderr please)

tests/test-vm.sh
```diff
@@ -72,17 +73,30 @@ declare -a qemu_command
...
 DPKG_ARCHITECTURE=$(dpkg --print-architecture)
 if [ "${DPKG_ARCHITECTURE}" = "amd64" ]; then
-    qemu_command=( qemu-system-x86_64 )
-    qemu_command+=( -machine q35 )
+  qemu_command=( qemu-system-x86_64 )
+  qemu_command+=( -machine q35 )
 elif [ "${DPKG_ARCHITECTURE}" = "arm64" ]; then
+  if [ "$RASPI" = 'yes' ]; then
+    if ! rpi_bootdata="$(sudo "$TEST_PWD"/tests/extract-rpi-bootdata.sh "$VM_IMAGE")"; then
+      echo "E: could not extract RPi boot data"
```
mika, line 81: redirect error message to stderr please

tests/test-vm.sh
```diff
@@ -28,7 +29,7 @@ bailout() {
     ps --pid="${QEMU_PID}" -o pid= | grep -q '.' && kill "${QEMU_PID:-}"
   fi
 
-  rm -rf "${TEST_TMPDIR}"
+  sudo rm -rf "${TEST_TMPDIR}"
```
mika, line 32: hm, only usage of `sudo` in this script here, is that expected?

tests/extract-rpi-bootdata.sh
```diff
@@ -0,0 +1,43 @@
...
+WORK_DIR="${WORK_DIR:-./rpi-work}"
+KERN_DIR="${KERN_DIR:-./rpi-kernel}"
+
+bail_noremove() {
+  echo "$1" >&2
+  exit 1
+}
+
+cleanup() {
+  if mountpoint "$WORK_DIR" 2>/dev/null >/dev/null; then
```
mika, line 12: that's quite unusual syntax for bash scripts, I'd prefer `>/dev/null 2>&1` or `&>/dev/null` instead (same below)

tests/docker-build-vm.sh
```diff
@@ -23,11 +24,17 @@ MIRROR='https://deb.debian.org/debian'
...
+  extra_buildopts='--vmfile'
+fi
+
 grml-debootstrap \
   --debug \
   --force \
-  --vmfile \
-  --vmsize 3G \
+  ${extra_buildopts} \
+  --imagesize 3G \
```
mika, line 37: Hm, is there a good reason why we couldn't continue using `--vmsize` and need to replace it with `--imagesize`? :thinking:

mika (commented):
Briefly went over the code except for `grml-debootstrap itself` (had to explicitly load the diff to even see it for code review, that was unexpected :-/). I want to clarify the `--imagesize` vs. `--vmsize` situation before looking further into the actual `grml-debootstrap` code changes.

What's definitely missing yet before we can submit this, are according doc changes in `grml-debootstrap.8.txt` :)

(Thanks for working on this!)

tests/docker-build-vm.sh
ArrayBolt3: We could keep using `--vmsize` for sure, but I wanted to switch to `--imagesize` since now that the argument can be used for more than just VM images, I marked `--vmsize` as deprecated. They mean the same thing, I figured using the newer syntax would be less confusing, especially since this build command might be building an RPi image, not strictly a VM image (though it will later boot that not-really-a-VM-image in a VM so I guess it doesn't really matter that much).

tests/test-vm.sh
ArrayBolt3: It is, the process of extracting the kernel and initramfs from the RPi image has to be done as root, so the kernel and initramfs end up being owned by root. The script will crash on this `rm -rf` if it isn't run with `sudo`. I did consider just changing the permissions on the kernel and initramfs files instead, but that was more work than this, in a context where I didn't think this had any security implications, so I opted for this intead.

mika: I see, thanks for clarifying! :+1:

ArrayBolt3:
While working on the docs, I'm realizing the `--vm` mode (which the `--rpi` mode is modeled after) automatically enables `--defaultinterfaces`. I think this makes a lot of sense for Raspberry Pi images. I'll try to sneak that in there before this is merged.

ArrayBolt3:
Tested the new default interfaces commit on my RPi4, works like a charm, don't need to do anything special for the network to work out of the box. CI passes and documentation is updated.

ArrayBolt3:
Hmm, that's a good point. I didn't realize environment variables were
a legal way to pass in parameters. That may require some rework.

ArrayBolt3:
Reworked, tested locally, `VMSIZE` and `IMAGESIZE` environment variables are being respected as appropriate now.

tests/docker-build-vm.sh
ArrayBolt3: Fixed the environment variable issue. Will leave this thread open so zeha can comment on the `--vmsize`/`--imagesize` stuff.

mika:
Thanks, LGTM. I think we should be good to apply this, though I'm afraid it's nothing we can get into Debian/trixie at this stage. To not block ourselves for further changes/fixes of grml-debootstrap that are supposed to go into Debian/trixie, we need to postpone this merge until the trixie freeze is gone.

Assigning this to our 2025/Q3 project/milestone accordingly, any further testing/feedback appreciated in the meanwhile! :)

zeha (approved):
from a glance it looks ok to me

PR #337, opened 2025-04-29
Set root partition type UUIDs for VM images according to the Discoverable Partition Specification

ArrayBolt3:
Pretty much what the title says, not very complicated. Worthy of note, this only affects VMs using GPT partitioning (since that's mandatory for partition type UUIDs), which currently means non-EFI systems won't get this fanciness due to the fact that grml-debootstrap doesn't support making BIOS-only GPT-formatted VMs.

Tested both amd64 and arm64 VM builds on a Trixie machine, the root partition type UUID is set properly in both instances.

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -1425,6 +1425,11 @@ prepare_vm() {
     arm64)      VMEFI=1;;
   esac
 
+  parted_type_command_supported='no'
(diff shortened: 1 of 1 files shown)


mika (approved):
I like it, LGTM && thx!
@zeha @jkirk ack?

zeha (approved):

Issue #339, opened 2025-05-09
Invalid install_devices configuration with grub-cloud-amd64


mika:
In commit 2891f1730f4673c979ac3b8dd0bae0088e7921cc (/cc @ArrayBolt3) we introduced the following new behavior:

```
if is_grub_bios_compatible && [ -n "$EFI" ]; then
MAIN_GRUB_PACKAGE="grub-cloud-amd64"
[...]
  echo "Setting grub debconf configuration for install device to $GRUB"
  if [ -n "$MAIN_GRUB_PACKAGE" ]; then
    echo "${MAIN_GRUB_PACKAGE} ${MAIN_GRUB_PACKAGE}/install_devices multiselect ${grub_device}" | debconf-set-selections
  fi
```

This ends up with:

```
root@grml:/# LC_ALL=C debconf-show grub-cloud-amd64  
* grub-cloud-amd64/install_devices: /dev/
```

But `grub-cloud-amd64` itself doesn't even support this setting, AFAICT.
It's supposed to be  `grub-pc/install_devices` only (the `grub-pc-bin` package is pulled in via `grub-cloud-amd64`).

This might break installations, when invoking it with `--grub /dev/` (yes, that's broken :)):

```
# grml-debootstrap --arch amd64 --grub /dev/ --filesystem ext4 --hostname foobar  --debopt '--variant=minbase --include=systemd,systemd-sysv,init,zstd,isc-dhcp-client,ifupdown,ca-certificates,qemu-guest-agent,eatmydata,usrmerge' --keep_src_list  --defaultinterfaces -r bookworm -t /dev/mapper/ngcp-root --efi /dev/nvme1n1p2 --password foobar 
 * EFI support detected.
 * grml-debootstrap [0.119] - Please recheck configuration before execution:

   Target:          /dev/mapper/ngcp-root
   Install grub:    /dev/
   Install efi:     /dev/nvme1n1p2
   Using release:   bookworm
[...]
Setting up mokutil (0.6.0-2) ...
Setting up libbrotli1:amd64 (1.0.9-2+b6) ...
Setting up gettext-base (0.21-12) ...
Setting up libfuse2:amd64 (2.9.9-6+b1) ...
Setting up libpng16-16:amd64 (1.6.39-2) ...
Setting up shim-signed-common (1.44~1+deb12u1+15.8-1~deb12u1) ...
No DKMS packages installed: not changing Secure Boot validation state.
[...]

PR #341, opened 2025-05-09
Don't set bogus debconf variable on grub-cloud packages

ArrayBolt3:
Fixes part of https://github.com/grml/grml-debootstrap/issues/339.

### DIFF
--- a/chroot-script
+++ b/chroot-script
@@ -728,8 +728,8 @@ grub_install() {
      grub_device="${GRUB}"
   fi
 
-  echo "Setting grub debconf configuration for install device to $GRUB"
-  if [ -n "$MAIN_GRUB_PACKAGE" ]; then
+  if [ -n "$MAIN_GRUB_PACKAGE" ] && ! [[ "$MAIN_GRUB_PACKAGE" =~ ^grub-cloud ]]; then
+    echo "Setting grub debconf configuration for install device to $GRUB"
     echo "${MAIN_GRUB_PACKAGE} ${MAIN_GRUB_PACKAGE}/install_devices multiselect ${grub_device}" | debconf-set-selections
   fi
 


zeha:
It might be necessary on the real packages, even when installing -cloud. Should be checked.

ArrayBolt3:
@zeha The real packages do not have this parameter either. The metapackages `grub-pc` and `grub-efi-ARCH` have this parameter, as they are the ones that usually handle reinstalling GRUB when updated. When using grub-cloud, neither of those packages are installed, and grub-cloud handles reinstalling GRUB when updated without using debconf variables.

mika:
> @zeha The real packages do not have this parameter either. The metapackages `grub-pc` and `grub-efi-ARCH` have this parameter, as they are the ones that usually handle reinstalling GRUB when updated. When using grub-cloud, neither of those packages are installed, and grub-cloud handles reinstalling GRUB when updated without using debconf variables.

Thanks for your PR and working on this!

Installing `grub-cloud-amd64` pulls in `grub-pc-bin` though.
But I'm wondering whether we should set `grub-pc/install_devices` no matter which package gets installed? :thinking:

ArrayBolt3:
> Installing grub-cloud-amd64 pulls in grub-pc-bin though.

`grub-pc` != `grub-pc-bin`. See `apt-cache show grub-pc` vs `apt-cache show grub-pc-bin`.

> But I'm wondering whether we should set grub-pc/install_devices no matter which package gets installed? 🤔

No. As warned by the debconf-set-selections manpage:

```
WARNING
       Only use this command to seed debconf values for packages that will be
       or are installed. Otherwise you can end up with values in the database
       for uninstalled packages that will not go away, or with worse problems
       involving shared values. It is recommended that this only be used to
       seed the database if the originating machine has an identical install.
```

zeha (approved):

zeha:
Okay, lets roll with this. I'm wondering if we should have an upgrade test of a non-bios-compatible EFI-grub install.

PR #343, opened 2025-05-27
Skip udevadm runs on Skip devices like

mika:
grml-debootstrap iterates over the present network devices by looking through /sys/class/net/, but file "bonding_masters" can't be queried:

```
| root@grml ~ # udevadm info --query=all --path=/sys/class/net/bonding_masters
| Unknown device "/sys/class/net/bonding_masters": No such device
```

As a result, execution of grml-debootstrap fails when bonds are used. Instead, let's ensure that any network interface we want to operate on is actually based on (a symlink pointing to) a directory which we can query via udevadm.

FTR:

```
| root@grml ~ # ls -l /sys/class/net
| total 0
| lrwxrwxrwx 1 root root    0 May 27 14:20 bond0 -> ../../devices/virtual/net/bond0
| -rw-r--r-- 1 root root 4096 May 27 14:20 bonding_masters
| lrwxrwxrwx 1 root root    0 May 27 14:20 eth0 -> ../../devices/pci0000:00/0000:00:12.0/virtio1/net/eth0
| lrwxrwxrwx 1 root root    0 May 27 14:20 eth1 -> ../../devices/pci0000:00/0000:00:13.0/virtio2/net/eth1
| lrwxrwxrwx 1 root root    0 May 27 14:20 eth2 -> ../../devices/pci0000:00/0000:00:14.0/virtio3/net/eth2
| lrwxrwxrwx 1 root root    0 May 27 14:20 lo -> ../../devices/virtual/net/lo
```

Thanks: @lexxua


zeha (approved):

mika:
Confirmed by @lexxua that this works, thanks a lot for your PR, feedback and testing! :bow:

Issue #344, opened 2025-05-28
remove kpartx workaround


adrelanos:
```
  kpartx -d "${ORIG_TARGET}" >/dev/null
  # Workaround for a bug in kpartx which doesn't clean up properly,
  # see Debian Bug #891077 and Github-PR grml/grml-debootstrap#112
  if dmsetup ls | grep -q "^${LOOP_PART} "; then
    kpartx -d "/dev/${LOOP_DISK}" >/dev/null
  fi
```

* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=891077
* https://github.com/grml/grml-debootstrap/pull/112

Should be fixed? Can this workaround be removed?

jkirk:
Ah, thanks for bringing this up.

I don't have the steps to reproduce for the problem with kpartx anymore (we worked on this 7 years ago at the Debconf18 in Taiwan. Brings back memories. 😊 ).

Did you test `grml-debootstrap` without that "workaround"? Did it work reliably?
But now that [#891077](mailto:891077@bugs.debian.org) is fixed, I think we should be safe to remove the workaround.

/cc @jkirk @zeha

adrelanos:
The mentioned Debian bug report states fixed indeed. However, there seems to be a regression.

https://github.com/grml/grml-debootstrap/issues/345 seems like a valid report.

The best outcome would be:

* short term: fix https://github.com/grml/grml-debootstrap/issues/345
* long term: report regression issue upstream.

(This bug report was from theoretic review of the source code. Also because I didn't experience the issue. I thought it's a good idea to ask if anyone knows more about it. A few days later, I learned that however this is still an issue in some situations, will elaborate in #345.)

Issue #345, opened 2025-05-30
Loop devices are not removed because grep fails


tabletseeker:
grep in `dmsetup ls | grep -q "^${LOOP_PART} "` fails because of this space `^${LOOP_PART} "`
https://github.com/grml/grml-debootstrap/blob/eae91811116fbc452efe45c6b80ecc647547ea1f/grml-debootstrap#L331
```
  if [ -n "${ORIG_TARGET}" ] ; then
    einfo "Removing loopback mount of file ${ORIG_TARGET}."
    kpartx -d "${ORIG_TARGET}" || eend $?
    # Workaround for a bug in kpartx which doesn't clean up properly,
    # see Debian Bug #891077 and Github-PR grml/grml-debootstrap#112
    if dmsetup ls | grep -q "^${LOOP_PART} "; then
      kpartx -d "/dev/${LOOP_DISK}" >/dev/null || eend $?
    fi
  fi
```
```
++ printf ' %s*%s Removing loopback mount of file /home/user/test.img.\n' '' ''
 * Removing loopback mount of file /home/user/test.img.
++ LAST_E_CMD=einfon
++ return 0
++ return 0
++ kpartx -d /home/user/test.img
++ dmsetup ls
++ grep -q '^loop2p3 '
++ '[' -n 1 ']'
++ EXIT=1
++ '[' -n '' ']'
++ exit 1
```
Removing the space allows grep to work correctly and `kpartx -d "/dev/${LOOP_DISK}"` to be executed.
```
    if dmsetup ls | grep -q "^${LOOP_PART}"; then
      kpartx -d "/dev/${LOOP_DISK}" >/dev/null || eend $?
    fi
```
Test with `/dev/loop2`
```
dmsetup ls | grep '^loop2p3'; echo $?
loop2p3	(252:8)
0
```

zeha:
Ok, but just removing the space is not correct - it would also match `loop22p3` (in your case) ...

zeha:
Related: #344 - which wants to remove this entirely

tabletseeker:
```
 * Removing loopback mount of file /home/user/test.img.
++ LAST_E_CMD=einfon
++ return 0
++ return 0
++ kpartx -d /home/user/test.img
++ dmsetup ls
++ grep -q '^loop0p3 '
++ '[' -n 1 ']'
++ EXIT=1
++ '[' -n '' ']'
++ exit 1
```

```
dmsetup ls
loop0p1	(252:0)
[...]

PR #346, opened 2025-06-02
Remove kpartx workaround

zeha:
Debian bug #891077 was fixed a long time ago, remove the workarounds.

See grml/grml-debootstrap#344.


adrelanos:
No.

https://github.com/grml/grml-debootstrap/issues/345 actually seems like a valid report.

So this PR would actually worsen things.

PR #347, opened 2025-06-25
Better isolate new chroot from host /dev, /dev/pts, /run/udev

zeha:
Closes: #1108311

Thanks: Wolfgang Zarre <lxdev12@zirdeon.com>


zeha:
I've asked around a bit and the consensus is: linux devtmpfs will not help us. What we probably need to do is create our own /dev, and use udev to populate it before installing the boot loader etc.

zeha:
Related: https://github.com/grml/grml-debootstrap/issues/348

the-linux-wolf:
Ok, joined now the party here.
Hmmm, very interesting, I cannot reproduce the mentioned failure above on my systems.
I bootstrapped several EFI based images with zeha/1108311 which are having a valid grub.cfg with correct UUID's and those are booting normal in qemu.
Seems to be rather another issue here, will have a look at it.

PR #378, opened 2026-07-16
Stop creating /etc/kernel-img.conf

zeha:
Compare https://salsa.debian.org/kernel-team/linux-base/-/commit/9977e3b63ba7d7f5bfcd2371505de245cf3e4000 .

- do_initrd is useless since 2016
- do_symlinks=yes is the default since 2016

Pointed out by cacin on IRC.


zeha:
installing goss fails in the ci jobs. dunno why.

mika (approved):

PR #380, opened 2026-07-23
Make FIXED_DISK_IDENTIFIERS="yes" work for GPT and Raspberry Pi images

ArrayBolt3:
If FIXED_DISK_IDENTIFIERS was set to "yes", a deterministic MBR disk ID would be set, but GPT disk and partition IDs were left as whatever Parted generated. It would also fail to create a deterministic ID for Raspberry Pi images. Fix both cases.

I've tested this when building VM images without EFI, VM images with EFI, and Raspberry Pi images. The first 2048 bytes of the images are fully reproducible when running grml-debootstrap with the same options across two image builds. (Note that this doesn't provide *full* reproducibility for the images yet; there is still substantial work to be done there, mainly with getting deterministic filesystem timestamps and the like, but that will take at least a while longer and can be done in the future.)

### DIFF
--- a/grml-debootstrap
+++ b/grml-debootstrap
@@ -50,7 +50,7 @@ MNTPOINT="/mnt/debootstrap.$$"
 [ -n "$DEBIAN_FRONTEND" ] || DEBIAN_FRONTEND='noninteractive'
 [ -n "$DEFAULT_LANGUAGE" ] || DEFAULT_LANGUAGE='en_US:en'
 [ -n "$DEFAULT_LOCALES" ] || DEFAULT_LOCALES='en_US.UTF-8'
-[ -n "$DISK_IDENTIFIER" ] || DISK_IDENTIFIER='26ada0c0-1165-4098-884d-aafd2220c2c6'
+[ -n "$FS_IDENTIFIER" ] || FS_IDENTIFIER='26ada0c0-1165-4098-884d-aafd2220c2c6'
 [ -n "$EXTRAPACKAGES" ] || EXTRAPACKAGES='yes'
 [ -n "$FALLBACK_MIRROR" ] || FALLBACK_MIRROR='http://deb.debian.org/debian'
 [ -n "$FIXED_DISK_IDENTIFIERS" ] || FIXED_DISK_IDENTIFIERS="no"
@@ -1311,8 +1311,8 @@ mkfs() {
         eerror "Not changing disk uuid for $TARGET because $MKFS doesn't seem to match for ext{2,3,4} file system"
         bailout 1
       else
-        einfo "Changing disk uuid for $TARGET to fixed (non-random) value $DISK_IDENTIFIER using tune2fs"
-        tune2fs "$TARGET" -U "$DISK_IDENTIFIER" </dev/null
+        einfo "Changing disk uuid for $TARGET to fixed (non-random) value $FS_IDENTIFIER using tune2fs"
+        tune2fs "$TARGET" -U "$FS_IDENTIFIER" </dev/null
       fi
     fi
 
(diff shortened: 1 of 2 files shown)


mika:
FYI: the failing CI runs are unrelated to this change but caused by goss upstream installation issue, see https://github.com/grml/grml-debootstrap/issues/381

mika (approved):
LGTM, would appreciate a second pair of eyes by @zeha

grml-debootstrap
```diff
@@ -50,7 +50,9 @@ MNTPOINT="/mnt/debootstrap.$$"
 [ -n "$DEBIAN_FRONTEND" ] || DEBIAN_FRONTEND='noninteractive'
 [ -n "$DEFAULT_LANGUAGE" ] || DEFAULT_LANGUAGE='en_US:en'
 [ -n "$DEFAULT_LOCALES" ] || DEFAULT_LOCALES='en_US.UTF-8'
-[ -n "$DISK_IDENTIFIER" ] || DISK_IDENTIFIER='26ada0c0-1165-4098-884d-aafd2220c2c6'
+[ -n "$FS_IDENTIFIER" ] || FS_IDENTIFIER='26ada0c0-1165-4098-884d-aafd2220c2c6'
+[ -n "$GPT_IDENTIFIER_PREFIX" ] || GPT_IDENTIFIER_PREFIX='b3170792-c1b6-4b84-b6ac'
```
zeha: wondering if we shouldn't use something more obviously non-random for both FS_... and GPT_... , like already done in MBR_IDENTIFIER.

ArrayBolt3: Arguably something pseudo-random is probably a good idea so that someone can place one disk provisioned with a reproducible image in a system with other existing disks without accidentally conflicting. (Of course, putting two reproducible disks in the same system is going to blow up, but we might be able to at least avert that in some situations.)

I don't know how FS_IDENTIFIER was generated, but I generated GPT_IDENTIFIER_PREFIX using `cat /proc/sys/kernel/random/uuid`.

ArrayBolt3: (Conversely, one could argue that a non-random-like UUID will warn the user that they really need to change the IDs to ensure uniqueness, since lack of uniqueness will cause issues in some situations. I don't have a strong opinion either way, I'm happy to switch if you'd like.)

zeha: IMO: either we do a pseudo-random identifier based on SOURCE_DATE_EPOCH, or we use something very clearly non-random. A random-looking yet non-random value is bad.

ArrayBolt3: I'm planning a future PR that uses SOURCE_DATE_EPOCH for making other FS bits reproducible, so I think the first option sounds like a good idea.

ArrayBolt3: @zeha Implemented SOURCE_DATE_EPOCH-based disk identifiers.

grml-debootstrap
```diff
@@ -739,6 +827,25 @@ fi
...
+  mbr_id_from_seed "${SOURCE_DATE_EPOCH}" MBR_IDENTIFIER
+  uuid_from_seed "${SOURCE_DATE_EPOCH}" 5 FS_IDENTIFIER
+  # The GPT partition identifer shouldn't be the similar to the FS identifier.
+  # To avoid confusion if multiple builds are kicked off in quick succession,
+  # it shouldn't be based on SOURCE_DATE_EPOCH + 1 either.
+  uuid_from_seed "$(( SOURCE_DATE_EPOCH ^ 0x0fffffffffffffff ))" 4 \
+    GPT_IDENTIFIER_PREFIX
+  printf '%s\n' "DEBUG: ${MBR_IDENTIFIER}"
+  printf '%s\n' "DEBUG: ${FS_IDENTIFIER}"
+  printf '%s\n' "DEBUG: ${GPT_IDENTIFIER_PREFIX}"
```
mika: should we wrap this within a check for `DEBUG="true"` to report this in debug mode only maybe?

ArrayBolt3: Uh, that was accidentally left there during development. I'd be in favor of just removing it entirely.

mika: Ah alright, ack then! :)

Issue #381, opened 2026-07-23
CI: broken goss installation


mika:
Our CI runs are broken due to misbehavior in goss installation (quoting from e.g. https://github.com/grml/grml-debootstrap/actions/runs/29309707092/job/87010771149#step:8:11):

```
+ curl -fsSL https://goss.rocks/install
++ pwd
+ GOSS_DST=/home/runner/work/grml-debootstrap/grml-debootstrap/tests
+ sh
Downloading https://github.com/goss-org/goss/releases/download/v0.4.9/goss_0.4.9_linux_x86_64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100     9  100     9    0     0    105      0 --:--:-- --:--:-- --:--:--   105

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Error: Process completed with exit code 2.
```

This used to be like this (quoting from https://github.com/grml/grml-debootstrap/actions/runs/28008505123/job/82896005667):

```
+ curl -fsSL https://goss.rocks/install
++ pwd
+ GOSS_DST=/home/runner/work/grml-debootstrap/grml-debootstrap/tests
+ sh
Downloading https://github.com/goss-org/goss/releases/download/v0.4.9/goss-linux-arm64
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0 13.0M    0 65536    0     0   626k      0  0:00:21 --:--:--  0:00:21  626k
100 13.0M  100 13.0M    0     0  77.1M      0 --:--:-- --:--:-- --:--:--  194M
Goss v0.4.9 has been installed to /home/runner/work/grml-debootstrap/grml-debootstrap/tests/goss
goss --version
goss version v0.4.9
Downloading https://raw.githubusercontent.com/goss-org/goss/master/extras/dgoss/dgoss
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

[...]

PR #382, opened 2026-07-23
tests: use manual install method for goss

zeha:
And pin goss to currently latest version 0.4.9.

Works around #381.


zeha:
Need to wait for Debian unstable to get back into an installable state before rerunning CI.

mika (approved):

PR #384, opened 2026-07-28
Add full support for reproducible image builds

ArrayBolt3:
There's still a lot of work to do on this (testing of various different installation scenarios and image types, adding CI tests, seeing if the same image can be generated by using the same parameters to grml-debootstrap on two *different* computers, etc.), but at this point. I'm able to build images back-to-back on the same system with the same parameters, and get two identical images as output. Getting this to work is tricky though, and there are a lot of caveats:

* `/etc/debootstrap/config` needs to have three settings set:
  * `KEEP_SRC_LIST=yes`
  * `MIRROR='/etc/debootstrap/etc/apt/sources.list.d/debian.sources` (`MIRROR` was never intended to take a file instead of a URL, but if one doesn't enable the backports repo and passes `KEEP_SRC_LIST='yes'`, the only place MIRROR ends up getting used is as an argument to mmdebstrap, and mmdebstrap *does* take a file as input for mirrors)
  * `SOURCE_DATE_EPOCH='...'` (set whatever time you want here)
* `/etc/debootstrap/etc/apt/sources.list.d/debian.sources` needs to be created with content similar to the following, to point it to whatever snapshots one wants from snapshot.debian.org:

```
Types: deb
URIs: https://snapshot.debian.org/archive/debian/20260727T202418Z/
Suites: trixie trixie-updates
Components: main contrib non-free non-free-firmware
Enabled: yes
Check-Valid-Until: no
Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg

Types: deb
URIs: https://snapshot.debian.org/archive/debian-security/20260726T121236Z/
Suites: trixie-security
Components: main contrib non-free non-free-firmware
Enabled: yes
Check-Valid-Until: no
Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
```

* You can't use the `--backports` option, as it triggers a code path in `chroot-script` that assumes `MIRROR` is a URL.
* You can't use the `--grml` option, as this adds a non-snapshotted grml repo to the image.
* You can't use filesystems other than ext4. (This will probably be a long-term limitation, since the steps to get a reproducible filesystem image will vary from filesystem to filesystem.)
* This probably won't provide reproducibility *yet* if installing directly to a hardware device, since it relies on `mkfs.ext4` being able to fstrim the target device during filesystem reimaging, which simply won't work on some hardware devices.
* There are probably subtle bugs I'm missing.

Despite all the above though, if you set up everything as described above and then use a command like `sudo grml-debootstrap --vm --vmefi --vmfile --vmsize 2G --target ./vm.img --release trixie --password x --hostname reproducible-debian --force` to generate a VM image, you will *probably* get a reproducible image. I'll continue working on this, but I figured now was a good time to submit a draft PR.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -36,6 +36,8 @@ Recommends:
  kpartx,
  parted,
  qemu-utils,
+ mtools,
+ python3,
 Description: installer for pure Debian
  Installation program for Debian, for bare-metal and virtual
  machines.
--- a/chroot-script
+++ b/chroot-script
@@ -387,7 +387,7 @@ rpi_cmdline() {
 reconfigure() {
   if [ -n "$RECONFIGURE" ] ; then
      for package in $RECONFIGURE ; do
-         if dpkg --list "$package" >/dev/null 2>&1 | grep -q '^ii' ; then
+         if dpkg --list "$package" 2>/dev/null | grep -q '^ii' ; then
            DEBIAN_FRONTEND=$DEBIAN_FRONTEND dpkg-reconfigure "$package" || \
            echo "Warning: $package does not exist, can not reconfigure it."
          fi
--- a/chroot-script
+++ b/chroot-script
@@ -405,14 +405,9 @@ passwords()
     return 0
   fi
 
-  CHPASSWD_OPTION=
-  if chpasswd --help 2>&1 | grep -q -- '-m,' ; then
-     CHPASSWD_OPTION='-m'
-  fi
-
   if [ -n "$ROOTPASSWORD" ] ; then
      # shellcheck disable=SC2086
-     echo root:"$ROOTPASSWORD" | chpasswd $CHPASSWD_OPTION
+     echo root:"$ROOTPASSWORD" | chpasswd
      export ROOTPASSWORD=''
   else
     a='1'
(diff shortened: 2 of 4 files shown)


ArrayBolt3:
Note that it might not be necessary to put anything under `/etc/debootstrap`; since mmdebstrap seems to install the mirror file you give it into the chroot, it should be possible to create the mirror file anywhere, point `--mirror` at it, and set `KEEP_SRC_LIST=yes` and `SOURCE_DATE_EPOCH=...` in the environment. That will probably also avoid having duplicate sources in the built image, which is one issue I ran into with my initial test images.

debian/control
```diff
@@ -29,6 +29,7 @@ Depends:
  gawk,
  kmod,
  mmdebstrap,
+ python3,
```
ArrayBolt3: I really do not like that I'm suggesting the addition of a new dependency here, but this was the best / only way I could figure out how to generate pseudorandom numbers from a seed reliably. Bash's `RANDOM` was wrapping when passed a 63-bit seed, Perl's `srand` (I think?) has the same issue. OpenSSL might be able to be abused as a random number generator, but I'm not sure that would be any better. Python is already present on a lot of Debian systems, so my hope is this isn't a big issue.

Edit: Downgraded this to a recommends (along with another dependency needed, `mtools`) and added an explicit check for it when `SOURCE_DATE_EPOCH` is set.

ArrayBolt3:
@zeha, @mika: Could I trouble one of you to look over the basic way this is being done and see if you like it? I'd like to finish polishing this to the point where it can be merged, but don't want to finish it only to find that the method being used is unsuitable.

PR #385, opened 2026-07-31
debian/control drop dependency gawk

crpb:
We do not need explicit gawk since commit 43cc740ba83a2613c7ec3ed3c16e4c549d0f2372


zeha (approved):

zeha:
NB: awk is pseudo-essential via a base-files Pre-Depends, so we should be good

